PR #37, opened 2015-08-25
Current user can now be retrieved via GET users/current.

### LINKED ISSUE #13: Current User implementation
/users/current should point to the current logged in user. Can be done with Spring Security

WouterKvG:
Call users/current with basic authentication (user:password or admin:password) and see the username. Fixes #13.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/service/UserService.java
+++ b/src/main/java/ch/wisv/areafiftylan/service/UserService.java
@@ -15,6 +15,8 @@ public interface UserService {
 
     Collection<User> getAllUsers();
 
+    String getCurrent();
+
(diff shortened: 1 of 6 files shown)


src/main/java/ch/wisv/areafiftylan/service/UserService.java
```diff
@@ -15,6 +15,8 @@
 
     Collection<User> getAllUsers();
 
+    String getCurrent();
```
skamoen: Waarom een String en niet gewoon de hele user?

skamoen:
Replaced by #44

Sorry @WouterKvG 
image

PR #43, opened 2015-09-26
Mail Service

skamoen:
You can make post requests to /mail

For example

POST /mail/users/1

{
"subject":"testmail", 
"message":"emptymessage"
}

image


src/main/java/ch/wisv/areafiftylan/Application.java
```diff
@@ -22,35 +22,35 @@
...
+                            User account = accountRepository.save(new User(a,
+                                    "password", a + "@mail.com"));
+                            User captain = accountRepository.findByUsername(a).get();
+                            Team team = new Team(a + "team", captain);
+                            teamRepository.save(team);
+
+
+                        });
+
+    }
```
nordinvannes: Application must be not be overwritten with this in master.

src/main/java/ch/wisv/areafiftylan/controller/MailRestController.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @RequestMapping(value = "/team/{teamId}", method = RequestMethod.POST)
+    ResponseEntity<?> sendMailToTeam(@PathVariable Long teamId, @Validated @RequestBody MailDTO mailDTO) {
+        Team team = teamService.getTeamById(teamId);
+        try {
+            mailService.sendTemplateMailToTeam(team, mailDTO);
+        } catch (MessagingException e) {
+            e.printStackTrace();
+        }
```
nordinvannes: With this code, you get the team from the service, and then push it back to the service with the mail. A better manner will be to send the teamId and the mail directly to the service.

Issue #6, opened 2015-08-05
Security: User restrictions on /teams endpoint


skamoen:
When viewing teams, users should only be able to see User.displayName

Issue #5, opened 2015-08-05
Security: User restrictions on /users endpoint


skamoen:
User should only be able to access their own user info. All other ID's should throw an unauthorized status.

PR #44, opened 2015-09-27
Spring Security implementation

### LINKED ISSUE #38: Implement multi-user authentication from db
http://www.mkyong.com/spring-security/spring-security-form-login-using-database/

skamoen:
Initial Spring Security implementation.

Still needs a fix to protect from CSRF

Testing loginform can be found in /static/loginform.html (localhost:8080/loginform.html)

No filteringrules apart from /users have been set. This should be in different pull requests. 

Fixes #38 and #13

Allows for #6 and #5 to be implemented


src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Autowired
+    private UserDetailsService userDetailsService;
+
+    /**
+     * This method is responsible for the main security configuration. The formlogin() section defines how to login.
+     * POST requests should be made to /login with a username and password field. Errors are redirected to /login?error.
+     * The logout section is similar.
+     *
+     * The last section is about permissions. Anything related to Login should is accessible for everyone. Use this
```
nordinvannes: Login should is accessible for everyone. ??

src/main/java/ch/wisv/areafiftylan/model/User.java
```diff
@@ -63,14 +68,20 @@ public Long getId() {
...
         this.passwordHash = passwordHash;
     }
 
+    @Override
+    public Collection<? extends GrantedAuthority> getAuthorities() {
+        return roles;
+    }
+
+    @Override
+    public String getPassword() {
```
nordinvannes: add jsonIgnore to not get it by the users

PR #47, opened 2015-09-29
WIP: Migration to PostgreSQL.

skamoen:
Migration to PostgreSQL. Needs more testing on proper persistence and schema update. It works though!

Check new application.properties.sample for settings.


skamoen:
Included in #52

Issue #16, opened 2015-08-05
JSON views


skamoen:
Implement suitable JSON views for each GET request

For example, when viewing /users/ it's not neccesary to see every profile. When GETting /users/{userId}, it can be useful to see the profile

This has many applications within the API

PR #59, opened 2015-11-15
Team update

### LINKED ISSUE #58: Information hiding on /teams endpoint
Currently, the teams endpoint returns every piece of information of all team members, including full profiles. Because this isn't just the current logged-on User, this information should be limited to need-to-know only.

skamoen:
Updated the Team-related classes to the latest standards for security and responseformat. 

Added an integration test for the /teams/ endpoint and implemented some fixes because of the outcomes of these tests. 

Also fixes #58, #16 and #6


src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
```diff
@@ -25,40 +35,164 @@ public TeamRestController(TeamService teamService) {
...
+     *
+     * @param teamId Id of the Team
+     *
+     * @return A list of the members of the Team with the given Id. Public information only.
+     */
+    @PreAuthorize("@currentUserServiceImpl.canAccessTeam(principal, #teamId)")
+    @JsonView(View.Public.class)
+    @RequestMapping(method = RequestMethod.GET, value = "/{teamId}/members")
+    public Set<User> getTeamMembersById(@PathVariable Long teamId) {
+        System.out.println("Calling normal method");
```
JeremybellEU: Left-over syso's or are they for logging purposes?

nordinvannes:
Mooi.

PR #60, opened 2015-11-15
Loginform

### LINKED ISSUE #56: Proper loginform
We need a default loginform (similar to Wordpress) so that we don't have to manually go to the extremely basic loginform.html

skamoen:
Instead of the extremely bare manual login, this adds a simple login form at the /login endpoint.

Fixes #56

image


nordinvannes:
Set LoginForm to the templates in application.properties

JeremybellEU:
LGTM

PR #76, opened 2015-12-18
Added CORS Config

### LINKED ISSUE #74: Enable CORS
To fix browser errors when making requests:

This might help, not sure if it's applicable for our case though. https://spring.io/guides/gs/rest-service-cors/

skamoen:
This allows Cross-origin resource sharing (CORS), which is pretty obvious for a REST API.

Please test before merge. 

Fixes #74


TimvdLippe:
Updated branch to use a Filter instead of the other configuration file, as for some reason this was not working. Ref: http://stackoverflow.com/questions/32319396/cors-with-spring-boot-and-angularjs-not-working

PR #78, opened 2015-12-19
Split UserRestController

### LINKED ISSUE #77: Split up mappings in UserRestController
The UserRestController currently consists of various mappings. To increase readibility and have a better split of responsibilities, the mappings should be put in a designated controllers. E.g.:

1. UserRestController <br>
Handles creation, updates and deletions of users
* CurrentUserRestController <br>
Should handle authentication of current user and apply updates accordingly
* UserProfileRestController <br>
Handle creation and updates of user profiles

skamoen:
The UserRestController is now split into 3 seperate controllers for CurrentUsers, Profiles en Users.

Also removed dedicated /profile GET endpoints since these are already include in the /users responses. 

Due to some PreAuthorize optimizations, I had to externalize the AccessDeniedExceptionHandler. This is actually a good thing, since it reduces code duplication. 

Fixes #77

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/controller/UserProfileRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/UserProfileRestController.java
@@ -1,0 +1,98 @@
+package ch.wisv.areafiftylan.controller;
+
+import ch.wisv.areafiftylan.dto.ProfileDTO;
+import ch.wisv.areafiftylan.model.Profile;
+import ch.wisv.areafiftylan.model.User;
+import ch.wisv.areafiftylan.service.UserService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.security.core.Authentication;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.*;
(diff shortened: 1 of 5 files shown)


TimvdLippe:
LGTM :+1:

src/main/java/ch/wisv/areafiftylan/controller/UserProfileRestController.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Add a profile to the current user. An empty profile is created when a user is created, so this method fills the
+     * existing fields
+     *
+     * @param input A representation of the profile
+     *
+     * @return The user with the new profile
+     */
+    @PreAuthorize("isAuthenticated()")
+    @RequestMapping(value = "/current/profile", method = RequestMethod.POST)
```
TimvdLippe: This is the only method not with a path of `{userId}/profile`. Since the profile field is in the user, we might want to remove this method and instead use the `PUT` method `replaceCurrentUser`

skamoen: I disagree. This means the user can edit their user details directly as well. That's not something we want. Profile data is freely adjustable though.

TimvdLippe: User can already edit their details directly right? `replaceCurrentUser` is accessible by the current user to update his/her details

skamoen: Good point. We don't want that either though.

PR #79, opened 2015-12-20
Enable CSRF Protection

### LINKED ISSUE #73: CSRF Protection
The csrf.disable() is easy for testing, but we should definitely enable this for anything public. This is done for a rest api by sending a new token with each request in the header. See http://www.codesandnotes.be/2015/02/05/spring-securitys-csrf-protection-for-rest-services-the-client-side-and-the-server-side/

This involves some client/frontend code as well, so this should be documented really well, perhaps with some ready-to-use javascript or something

skamoen:
This enables CSRF protection. This means that every POST requests needs a CSRF token in the header at X-CSRF-TOKEN or your request will be denied. 

You can get a token from a GET on /login or /token

Fixes #73

Also, FUCK REST-Assured. Had to rewrite all the tests. 

/cc @stijncr If you manage to find time to write tests, make sure you do it in this new format.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
@@ -12,6 +12,9 @@ import org.springframework.security.config.annotation.web.configuration.WebSecur
 import org.springframework.security.core.userdetails.UserDetailsService;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.web.csrf.CsrfFilter;
+import org.springframework.security.web.csrf.CsrfToken;
+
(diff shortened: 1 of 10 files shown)


src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
```diff
@@ -23,16 +25,19 @@
      * This method is responsible for the main security configuration. The formlogin() section defines how to login.
      * POST requests should be made to /login with a username and password field. Errors are redirected to /login?error.
      * The logout section is similar.
-     *
-     * The last section is about permissions. Anything related to Login is accessible for everyone. Use this
-     * for URL-based permissions if that's the best way. Use Method specific permissions if this is not feasible.
-     *
+     * <p>
```
bvanderdrift: Why <p>?

Issue #68, opened 2015-11-18
Ticket/Order Controller & Service


skamoen:
A Controller on some endpoint (/tickets or /orders or something else) that handles requests related to orders and tickets

see #10, #65 , #66 , #67 for details and dependencies

skamoen:
Add discount code for early bird tickets, or another way of making tickets cheaper for early birds

skamoen:
Fixed in #86

Issue #67, opened 2015-11-18
Payment Service


skamoen:
A payment service that allows for multiple payment provider implementations

Check the Mollie API https://www.mollie.com/nl/docs/overzicht for some inspiration of required methods.

skamoen:
There's a new 3rd party Java module for Mollie. Looks basic but functional enough!

https://github.com/stil4m/mollie-api/

skamoen:
Fixed in #86

Issue #66, opened 2015-11-18
Order model


skamoen:
We need a model for the Order object

- [x] Contains a single or multiple tickets
- [x] Is linked to a single User
- [x] A single User can have multiple Orders
- [x] A price, based on the contained Tickets
- [x] A status (paid, cancelled ect)

TimvdLippe:
[Comment on old branch](https://github.com/AreaFiftyLAN/lancie-api/commit/d99095e0f0541b207269bc145627a60329910c5b#commitcomment-15080724)

skamoen:
Fixed in #86

PR #86, opened 2015-12-27
WIP: Ticketsale

### LINKED ISSUE #65: Ticket Interface or Abstract class
We need a Ticket interface or abstract class for the various types of tickets we want to be able to sell.

- [ ] It needs to be an interface that supports implementations for Event Tickets, Food Tickets, Pickupservice Tickets, ect.
- [x] It needs a price
- [ ] It needs a method to "enable/validate" the ticket.
- [ ] It needs a method to lock/unlock the ticket for transfer

skamoen:
Based on the CSRF Branch (#79)

Enable the Ticketsale in the backend. In this PR, it becomes possible to create orders, add tickets to orders, and checkout the order and pay them through Mollie. Still contains some things to do that are dependent on other PR's, so merge after #78 and #79 

Fixes #65, #66, #67, #68 

- [x] Merge #78 
- [x] Merge #79 
- [x] Include pricing in tests


src/main/java/ch/wisv/areafiftylan/model/Ticket.java
```diff
@@ -0,0 +1,103 @@
...
+    public void setOwner(User owner) {
+        this.owner = owner;
+    }
+
+    public TicketType getType() {
+        return type;
+    }
+
+    public float getPrice() {
+        return pickupService ? type.getPrice() + 5 : type.getPrice();
```
bvanderdrift: Not ok. Lets think about a solution. Worst case: constant on top of the class.

PR #91, opened 2015-12-28
Replaced hardcoded Urls for property based values

### LINKED ISSUE #51: Specify front-end URLs in config
For now, the mailservce does a GET request directly to the API. On launch, we need a way to make a GET to the frontend, which should in turn make the call to the API.

This could be set from a config properties file to make it easily changable.

skamoen:
Fixes #51

### DIFF
--- a/src/main/resources/config/application-test.properties
+++ b/src/main/resources/config/application-test.properties
@@ -23,3 +23,9 @@ spring.thymeleaf.suffix=.html
 spring.thymeleaf.mode=HTML5
 spring.thymeleaf.encoding=UTF-8
 spring.thymeleaf.content-type=text/html
+
+##AREA FIFTYLAN SETTINGS
+
+a5l.paymentReturnUrl=https://areafiftylan.nl/ordersuccess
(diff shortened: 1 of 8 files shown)


src/main/resources/config/application-test.properties
```diff
@@ -22,4 +22,10 @@ spring.thymeleaf.view-names= *
 spring.thymeleaf.suffix=.html
 spring.thymeleaf.mode=HTML5
 spring.thymeleaf.encoding=UTF-8
-spring.thymeleaf.content-type=text/html
\ No newline at end of file
+spring.thymeleaf.content-type=text/html
+
+##AREA FIFTYLAN SETTINGS
+
+a5l.paymentReturnUrl=https://areafiftylan.nl/ordersuccess
```
TimvdLippe: This will break if the server is run on a different path, e.g. `v1/api/`. Would it be possible to base the url on the root path or must it be manually updated?

skamoen: This is meant as a front-end URL. You can't redirect a user to the api ;). The front-end would make a call to the API.

Issue #81, opened 2015-12-22
Order clean task


skamoen:
The application should check every few minutes for unpaid Orders older then $expired. This Order should be disabled and Tickets must be removed so another user can order a ticket of this type again.

See https://spring.io/guides/gs/scheduling-tasks/ for a guide on scheduling these things.

skamoen:
Fixed in #92

PR #92, opened 2015-12-28
Added Order Expiry functionality to API.

bvanderdrift:
Solves #81 

Expiry Time values (Minutes that a order can stay open, how often this is checked) are hardcoded into the ScheduledTasksService class.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/service/ScheduledTasksService.java
+++ b/src/main/java/ch/wisv/areafiftylan/service/ScheduledTasksService.java
@@ -1,0 +1,41 @@
+package ch.wisv.areafiftylan.service;
+
+import ch.wisv.areafiftylan.model.ExpiredOrder;
+import ch.wisv.areafiftylan.model.Order;
+import ch.wisv.areafiftylan.service.repository.ExpiredOrderRepository;
+import ch.wisv.areafiftylan.service.repository.OrderRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+import java.time.LocalDateTime;
+import java.util.Collection;
+
+/**
+ * Created by beer on 28-12-15.
+ *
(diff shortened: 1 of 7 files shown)


src/main/java/ch/wisv/areafiftylan/service/repository/OrderRepository.java
```diff
@@ -4,11 +4,14 @@
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.time.LocalDateTime;
 import java.util.Collection;
+import java.util.Date;
```
skamoen: I don't think you're using util.Date

src/main/java/ch/wisv/areafiftylan/service/ScheduledTasksService.java
```diff
@@ -0,0 +1,41 @@
...
+import java.time.LocalDateTime;
+import java.util.Collection;
+
+/**
+ * Created by beer on 28-12-15.
+ *
+ * Class with all scheduled tasks
+ */
+@Component
+public class ScheduledTasksService {
```
skamoen: I'd rename this to something like TaskScheduler

src/main/java/ch/wisv/areafiftylan/service/ScheduledTasksService.java
```diff
@@ -0,0 +1,41 @@
...
+    private final int ORDER_EXPIRY_CHECK_INTERVAL_SECONDS = 60;
+
+    @Autowired
+    private OrderRepository orderRepository;
+
+    @Autowired
+    private ExpiredOrderRepository expiredOrderRepository;
+
+    @Autowired
+    private OrderService orderService;
```
skamoen: With 3 @Autowired fields, put these in a single Constructor

```
@Autowired
public ScheduledTasksService(OrderRepository orderRepository, ExpiredOrderRepository expiredOrderRepository,OrderService orderService){ ... }
```

PR #94, opened 2015-12-31
Add User tests, better Validation and small fixes

skamoen:


src/main/java/ch/wisv/areafiftylan/dto/ProfileDTO.java
```diff
@@ -8,16 +8,16 @@
 public class ProfileDTO {
```
TimvdLippe: Lombok would be sooo nice for this class. No more manual generated setters.

src/test/java/ch/wisv/areafiftylan/UserRestIntegrationTest.java
```diff
@@ -1,17 +1,91 @@
 package ch.wisv.areafiftylan;
```
TimvdLippe: This class is very very big, while it also contains tests for the split-up controllers. I would split the tests according to the changes made in #78

TimvdLippe:
We might want to consider using Lombok for the DTO's. Other than that 1 comment regarding the `UserRestIntegrationTest`, then LGTM :+1:

src/main/java/ch/wisv/areafiftylan/dto/ProfileDTO.java
skamoen: See #95

src/test/java/ch/wisv/areafiftylan/UserRestIntegrationTest.java
skamoen: It's a test class, testing all aspect of Users. What would we gain by splitting it up except for a couple of seconds when scrolling through? I don't think it's worth the extra files and lines in your debugger when running them.

Issue #90, opened 2015-12-28
Duplicate ticket display on GET /orders/{id}


skamoen:
When making a GET request on a specific order, sometimes you see the same ticket (Same ID) twice. Not good.

- [x] Fixed
- [x] Checked in tests

skamoen:
Fixed in #96

PR #96, opened 2015-12-31
Add Order Id to response

TimvdLippe:
The Id is needed in the frontend to post. This is easier than the location, because the iron-ajax only needs the id and not the full location.


skamoen:
Fixes #90 now too

PR #98, opened 2016-01-02
Add checks for unset properties on application startup

TimvdLippe:
Implement fail-fast approach for custom properties on application startup.

### DIFF
--- a/src/main/resources/config/application.properties.sample
+++ b/src/main/resources/config/application.properties.sample
@@ -38,11 +38,11 @@ server.port=9000
 
 
 ## AREA FIFTYLAN SETTINGS
-# Key for sending an e-mailbody to the Mollie client
+# Key for sending a paymentrequest to the Mollie api
 #a5l.molliekey=
 # The return url for the frontend that displays the success message
 a5l.paymentReturnUrl=https://areafiftylan.nl/ordersuccess
 # The return url for the frontend that confirms a registration
 a5l.mail.confirmUrl=https://areafiftylan.nl/confirmRegistration
-# E-mail address used for sending e-mails using Mollie
+# E-mail address used for sending e-mails
 a5l.mail.sender=LANcie <lancie@ch.tudelft.nl>
--- a/src/main/resources/config/application.properties.sample
+++ b/src/main/resources/config/application.properties.sample
@@ -19,6 +19,7 @@ spring.datasource.driverClassName=org.postgresql.Driver
 # Change <MyDB> with the name of your database name
 #spring.datasource.url=jdbc:postgresql://localhost:5432/<MyDB>
 # Username to log in to the database
+# It's recommended to create a dedicated AreaFiftyLAN Database user instead of using the default postgres user
 #spring.datasource.username=<dbUser>
 # Password to log in with <dbUser> in database <MyDB>
 #spring.datasource.password=<dbUserPassword>
--- a/src/main/resources/config/application.properties.sample
+++ b/src/main/resources/config/application.properties.sample
@@ -1,9 +1,9 @@
 # Basic auth
 security.user.password =
 
-#Default Property file, which is committed in Git
-spring.mail.host=
-spring.mail.port=
+# Retriever for sending e-mails. Used for verify created users
+#spring.mail.host=
+#spring.mail.port=
 
 # THYMELEAF (ThymeleafAutoConfiguration)
 spring.thymeleaf.check-template-location=true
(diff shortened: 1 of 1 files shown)


src/main/java/ch/wisv/areafiftylan/service/MailServiceImpl.java
```diff
@@ -27,7 +28,14 @@
     String sender;
 
     @Autowired
-    public MailServiceImpl(JavaMailSender mailSender, SpringTemplateEngine templateEngine) {
+    public MailServiceImpl(JavaMailSenderImpl mailSender, SpringTemplateEngine templateEngine) {
+        if (mailSender.getHost() == null || mailSender.getHost().isEmpty()) {
+            throw new NullPointerException("Host for mailserver is not set.");
```
skamoen: I dont think Nullpointer is the right exception here...

src/main/java/ch/wisv/areafiftylan/service/MolliePaymentService.java
```diff
@@ -31,17 +31,27 @@
...
 
-    @Value("${a5l.paymentReturnUrl}")
     String returnUrl;
 
     @Autowired
-    public MolliePaymentService(OrderRepository orderRepository, @Value("${a5l.molliekey}") String apiKey) {
+    public MolliePaymentService(OrderRepository orderRepository,
+                                @Value("${a5l.molliekey}") String apiKey,
+                                @Value("${a5l.paymentReturnUrl}") String returnUrl) {
+        if (apiKey == null || apiKey.isEmpty()) {
```
skamoen: Gebruik hier liever Strings.isNullOrEmpty()

skamoen:
I dont think this is something we should check/fix in this way. The values are currently in the properties files so you can run them without the (private!) keys. If you remove these empty properties completely the Application wont launch, with appropriate exceptions.

TimvdLippe:
This PR addresses future problems with the maintainance of the application. New developers will need to configure the server using the properties file. The current sample file lists the properties. This sample file leaves all unknown properties empty, which results in Strings being empty.

By explicitly declaring that certain properties are required to use the application, future configuration problems can be prevented. This is especially important since next year the team will different.

Therefore I am convinced we should make the implicit requirements explicit, to prevent future mistakes and trouble.

praseodym:
Spring should throw an error if an `@Autowired` dependency is not marked `required=false` or a fallback (default) value is not defined for an `@Value`.

skamoen:
I think the best way to fix this, is to change the sample properties and make the a5l properties commented. That way the application will throw exceptions when these proeprties aren't uncommented and set. Perhaps add some documentation in the properties file as well

src/main/resources/config/application.properties.sample
```diff
@@ -14,24 +14,34 @@ spring.thymeleaf.mode=HTML5
 spring.thymeleaf.encoding=UTF-8
 spring.thymeleaf.content-type=text/html
```
skamoen: This comment applies to ` spring.thymeleaf.view-names= MailTemplate, loginForm`. This line can be removed as well (it only makes stuff complicated). I already did this in the neon-admin branch, but I guess you can do this in this PR before that is merged.

src/main/resources/config/application.properties.sample
```diff
@@ -14,24 +14,34 @@ spring.thymeleaf.mode=HTML5
...
 
-##DATABASE PROPERTIES
+## DATABASE PROPERTIES
 spring.datasource.driverClassName=org.postgresql.Driver
-spring.datasource.url=jdbc:postgresql://localhost:5432/<MyDB>
-spring.datasource.username=<dbUser>
-spring.datasource.password=<dbUserPassword>
+# Change <MyDB> with the name of your database name
+#spring.datasource.url=jdbc:postgresql://localhost:5432/<MyDB>
+# Username to log in to the database
```
skamoen: Perhaps add "It's recommended to create a dedicated AreaFiftyLAN Database user instead of using the default postgres user"

skamoen:
LGTM, if you can fix the two comments it's ready for merge

src/main/resources/config/application.properties.sample
TimvdLippe: What do you refer to with "this comment"?

src/main/resources/config/application.properties.sample
```diff
@@ -14,24 +14,35 @@ spring.thymeleaf.mode=HTML5
...
-server.port =
 
+# Default port used by frontend
+server.port=9000
 
-##AREA FIFTYLAN SETTINGS
 
-a5l.molliekey=
+## AREA FIFTYLAN SETTINGS
+# Key for sending an e-mailbody to the Mollie client
```
skamoen: Not sure what email has to do with Mollie. This is the authentication key for using the Mollie API

src/main/resources/config/application.properties.sample
```diff
@@ -14,24 +14,35 @@ spring.thymeleaf.mode=HTML5
...
+#a5l.molliekey=
+# The return url for the frontend that displays the success message
 a5l.paymentReturnUrl=https://areafiftylan.nl/ordersuccess
+# The return url for the frontend that confirms a registration
 a5l.mail.confirmUrl=https://areafiftylan.nl/confirmRegistration
-a5l.mail.sender=LANcie <lancie@ch.tudelft.nl>
\ No newline at end of file
+# E-mail address used for sending e-mails using Mollie
```
skamoen: Same here, email and Mollie are two completely different things :stuck_out_tongue:

PR #100, opened 2016-01-03
Adopt usage of lombok

martijn9612:


TimvdLippe:
LGTM :+1:

src/main/java/ch/wisv/areafiftylan/web/controller/WebRestController.java
```diff
@@ -21,13 +20,15 @@
     TournamentServiceImpl tournamentService;
     EventServiceImpl eventService;
     SponsorServiceImpl sponsorService;
+    CommitteeMemberServiceImpl committeeService;
```
skamoen: This is not how you're supposed to use Services in the controller. Services have an interface, and an implementation and you declare the interface as a field in your controller. A service for each web entity might be a little overkill, zo maybe combine these into two or three services (Something like EventService for tournaments, events and other... events) and a Webservice for other static content such as sponsors, committees and such.

skamoen:
The Lombok part looks good, but I would make a seperate PR for the web part.

martijn9612:
I have made a mistake apparently, the first commit should not be in this pull request. My bad. I will remove the pull request and open a correct one with the issues fixed.

PR #102, opened 2016-01-05
Adminpanel for Ticketsale

skamoen:
This is the initial admin panel. It allows viewing of registered users and orders. It can also manually update orders.

Depends on #101, rebase after merge.


Icety:
Looks good, maybe remove the unused links from the dashboard-index file.
Like languages, profile, mailbox and tasks.

PR #130, opened 2016-02-21
Removed late bird ticket

Icety:


TimvdLippe:
Do not comment out unused code, but rather delete it.

martijn9612:
Looks good to me, tests are all fixed too. Thanks!

Issue #134, opened 2016-02-22
Clean /static/dashboard assets


skamoen:
Currently the entire neon-dashboard assets folder is in the repo. This can probably be cleaned up a little.

Icety:
Dashboard needs to be redone, so the static data should be deleted completely.

PR #135, opened 2016-02-22
Add dashboard assets

skamoen:
This adds the whole neon-admin package. Overkill, but the dashboard wont function without it. See #134


skamoen:
@Icety Do you want to clean this up before merge, or do that later? I think we can save at least 20MB here. See #134

Icety:
I have no time right now sadly, so I would like to do it later.

skamoen:
In that case I'm just merging this, because the dashboard won't function without it

PR #138, opened 2016-02-23
Fix email urls

martijn9612:
Resolves the missing hashbang issue.


skamoen:
Why did you use A5L.user instead of  .mail?

martijn9612:
It made more sense to name it that way, since the other name was awfully generic.

skamoen:
We probably have to restructure the property  names anyway sometime in the future, so I'll just merge this :-)

PR #145, opened 2016-02-27
Move static folder to classpath to be included in jar

skamoen:


martijn9612:
Only moved some files to be included in the jar. Looks good to me.

PR #146, opened 2016-02-28
Remove redirects on login

skamoen:
Due to the way the app is running behind an Apache server, internal redirects fail miserably with mixed content errors. This fixes that behaviour by eliminating redirects. Instead, it returns a 200 OK for successful logins, and a 401 for unauthenticated requests and failed login attempts.


TimvdLippe:
Confirmed working. LGTM

PR #153, opened 2016-03-05
Update README.md

skamoen:
Stuff is about to change, but this updates the extremely outdated README.md to something good enough to represent the current state.


TimvdLippe:
LGTM

PR #154, opened 2016-03-05
Enable manual Order status updates

skamoen:
This PR adds a call to retreive the latest Order status at `/orders/{orderId}/status`. Next to that, the number of statusses has been updated to allow for non-pushable statusses coming from Mollie. The Order cleanup task has been expanded to include CANCELLED orders as well.


TimvdLippe:
Confirmed working together with https://github.com/AreaFiftyLAN/lancie-frontend/pull/120

Issue #159, opened 2016-03-08
Team invites


skamoen:
Currently, Team Captains can simply add members to their team by username. This should be a little more controlled. Team Captains can send invites to join a team, but a member only joins a team after accepting this invite. This can be implemented using a subclass of Token.

PR #169, opened 2016-03-11
Seatmap Update

### LINKED ISSUE #11: Seat extension
Seats need more functionality. Should have a Label and several configuration options. (add rows, ect)

Team captains should be able to reserve seats for their entire team

### LINKED ISSUE #162: Seat reserverations
In addition to the current seatmap branch, a commonly used/requested feature is the ability for admins to reserve ranges of seats for large groups. For this, some sort of AvailableState could be introduced, which could differentiate between Open, Taken, or Reserved. 
Taken seats would have an owner, while reserved seat wouldn't, but still won't be able to select by other users then admins. 

The isAvailable boolean could be reused for this and possible integrated into this AvailableState enum, as a seat would still be available, or not, independent on what sort of state the seat is in.

skamoen:
Big seatmap update to make it functional

- Each Ticket allows for the reservation of a seat
- Teamcaptains have "control" over reservations of team members
- Seats have a groupname, which can be accessed seperately `seats/{groupname}, and a number in said group. 
- A User can find his group at `/users/current/seat`
- You can look up all seats belonging to a team at `/teams/{teamName}/seats`

Fixes #162, Fixes #11

#### Future work
- Create an endpoint that lists tickets you have control over


TimvdLippe:
I have added some test code to verify the functionality with the frontend. A brief testing session was working, but most of the end-points I could not test yet because the code does not exist yet. We will need to adjust later if problems arise during that implementation. LGTM

src/main/java/ch/wisv/areafiftylan/controller/CurrentUserRestController.java
```diff
@@ -126,4 +130,11 @@ public Order getOpenOrder(Authentication auth) {
         UserDetails currentUser = (UserDetails) auth.getPrincipal();
         return orderService.getOpenOrder(currentUser.getUsername());
     }
+
+    @RequestMapping(value = "/seat", method = RequestMethod.GET)
+    public Seat getCurrentUserSeat(Authentication auth) {
```
TimvdLippe: We decided that seats are linked to tickets. How would this work if 1 user has multiple tickets?

src/main/java/ch/wisv/areafiftylan/controller/SeatRestController.java
```diff
@@ -26,22 +32,166 @@ public SeatRestController(SeatService seatService, UserService userService) {
...
+    /**
+     * Get the Seat of a specific User
+     *
+     * @param userId Id of the User you want the Seat of.
+     *
+     * @return The Seat of the given User.
+     */
+    @PreAuthorize("hasRole('ADMIN')")
     @RequestMapping(value = "/users/{userId}/seat", method = RequestMethod.GET)
     public Seat getSeatByUser(@PathVariable Long userId) {
```
TimvdLippe: Same as above comment, should probably be `List<Seat>`.

src/main/java/ch/wisv/areafiftylan/model/Seat.java
```diff
@@ -33,17 +45,30 @@ public void setTaken(boolean taken) {
...
+
+    public void setTicket(Ticket ticket) {
+        this.ticket = ticket;
+
+        this.taken = this.ticket != null;
     }
 
-    public void setUser(User user) {
-        this.user = user;
+    public String getSeatGroup() {
```
TimvdLippe: Could be replaced with `@Setter` and `@Getter`?

src/main/java/ch/wisv/areafiftylan/model/User.java
```diff
@@ -139,4 +139,33 @@ public void setAccountNonLocked(boolean accountNonLocked) {
     public void setEnabled(boolean enabled) {
         this.enabled = enabled;
     }
+
+    @Override
```
TimvdLippe: User can be annotated with `@EqualsAndHashCode` to generate these automatically.

src/main/java/ch/wisv/areafiftylan/service/CurrentUserServiceImpl.java
```diff
@@ -17,6 +20,10 @@
     @Autowired
     OrderService orderService;
 
+    // TODO: Move this to the future TicketService
```
TimvdLippe: Can this TODO still be fixed in this PR?

skamoen: No, this is coming in the Ticket Transfer PR.

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
```diff
@@ -1,43 +1,124 @@
...
 
     @Override
-    public List<Seat> getAllSeats() {
-        return seatRepository.findAll();
+    public SeatmapResponse getAllSeats() {
+        List<Seat> all = seatRespository.findAll();
+
+        Map<String, List<Seat>> seatMapResponse = new HashMap<>();
+
+        for (Seat seat : all) {
```
TimvdLippe: I think a solution with streams is easier to grasp. Something along the lines of

```java
all.stream().map(Seat::getSeatGroup).forEach(g -> seatMapResponse.put(g, new ArrayList<>()));
all.stream().forEach(s -> seatMapResponse.get(s.getSeatGroup).add(s);
```

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
```diff
@@ -1,43 +1,124 @@
...
+        seatMapResponse.put(groupname, seatGroup);
+
+        return new SeatmapResponse(seatMapResponse);
+    }
+
+    @Override
+    public boolean reserveSeatForTicket(String groupname, int seatnumber, Long ticketId) {
+        Ticket ticket = ticketRepository.findOne(ticketId);
+        Seat seat = seatRespository.findBySeatGroupAndSeatNumber(groupname, seatnumber);
+        if (!seat.isTaken()) {
```
TimvdLippe: Maybe this method should be `synchronized` to prevent two people from reserving the same seat in a race condition.

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
```diff
@@ -1,43 +1,124 @@
...
+        }
     }
 
     @Override
-    public Seat getSeatByCoordinate(Coordinate coordinate) {
-        return seatRepository.findByCoordinate(coordinate);
+    public boolean reserveSeat(String groupname, int seatnumber) {
+        Seat seat = seatRespository.findBySeatGroupAndSeatNumber(groupname, seatnumber);
+
+        if (!seat.isTaken()) {
```
TimvdLippe: Same as above comment.

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
```diff
@@ -1,43 +1,124 @@
...
+
+    @Override
+    public Set<Seat> getSeatsByTeamName(String teamName) {
+        Set<Seat> seats = new HashSet<>();
+        Team team = teamService.getTeamByTeamname(teamName).orElseThrow(() -> new TeamNotFoundException(teamName));
+
+        for (User user : team.getMembers()) {
+            Optional<Seat> seat = seatRespository.findByTicketOwnerUsername(user.getUsername());
+            if (seat.isPresent()) {
+                seats.add(seat.get());
```
TimvdLippe: I think these lines can be replaced with

```java
seatRespository.findByTicketOwnerUsername(user.getUsername()).ifPresent(seats::add);
```

See: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-

TimvdLippe:
Build is failing due to my changes, presumably because tests do not reset state.

Some methods return `Seat` instead of `List<Seat>` per discussion yesterday.

Other than that some Stream magic to make the code less verbose.

LGTM

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
JWGmeligMeyling: you can even do:

```java
return team.getMembers().steam()
    .map(User::getUsername)
    .map(seatRepository::findByTicketOwnerUsername)
    .filter(Optional::isPresent)
    .collect(Collectors.toSet());
```

Matter of preference. A bit sad that it is required to see whether the optional is present, in most stream implementations you get that for free.

src/main/java/ch/wisv/areafiftylan/service/TeamService.java
```diff
@@ -11,7 +11,7 @@
 
     Team save(Team team);
 
-    Optional<Team> getTeamById(Long id);
+    Team getTeamById(Long id);
```
JWGmeligMeyling: Seems inconsistent with `TeamService#getTeamByTeamname` to me?

src/main/java/ch/wisv/areafiftylan/dto/SeatGroupDTO.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.hibernate.validator.constraints.NotEmpty;
+
+import javax.validation.constraints.NotNull;
+
+/**
+ * Created by sille on 10-1-16.
+ */
+public class SeatGroupDTO {
```
JWGmeligMeyling: `@Getter` and `@Setter` can be omitted if you add `@Data` to the class, which also gives to `#toString()`, `#equals()` and `#hashcode` for free (which is nice in collections and test matchers).

src/main/java/ch/wisv/areafiftylan/model/Seat.java
JWGmeligMeyling: As Lombok is on the classpath this could indeed be replaced with annotations.

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
JWGmeligMeyling: Not sure if services in Spring are singletons by default. Otherwise, you need to synchronise manually on a singleton/static lock object. (Which is also a better practise over using the synchronized keyword)

src/main/java/ch/wisv/areafiftylan/service/TeamService.java
skamoen: This is fixed in #159

PR #172, opened 2016-03-15
Changed open orders to List-based representation

### LINKED ISSUE #131: Return other error message than 400
When an ajax calls generates errors, for instance when a field is not correctly entered in a form, the api returns a 400 error. The browser however throws an error in the console when this happens, even when it actually returns valid information.

Since browsers just want a 200 message, we should send this back, even when errors occur. Polymer should what to do based on the data that is returned and then show an error.

skamoen:
The Open orders of a User (which can still only be one) is now changed to a list based representation. This prevents the 404 error in the browser when there are no open orders. 

Fixes #131


TimvdLippe:
LGTM

PR #175, opened 2016-03-17
Add reference field to User response

skamoen:
This PR adds a reference field to the JSON response. It doesn't store anything in the database, but calculates a hash of the username field. Useful for front-end comparing logic. You can't actually do anything with it.


TimvdLippe:
LGTM

PR #176, opened 2016-03-22
Prevent test-data insertion when running tests

skamoen:
This PR prevents the CommandlineRunner in Application.class from running when running tests. This removes the need to clear all repositories first, so they're not a dependency of the base integrationTest class anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ crashlytics.properties
 crashlytics-build.properties
 
 application.properties
+application-dev.properties
 
 node_modules
 dist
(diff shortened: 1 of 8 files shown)


TimvdLippe:
Could this easily be extended for the dev-profile? I imagine something along the lines of `@Profile("dev")`

skamoen:
Yes. You only have to set your Active Profile somewhere for that to work, but it works the same way.

TimvdLippe:
I think it is better to do that instead since the tests then automatically do not run this method.

skamoen:
This basically does the same thing, as tests are always run as with the "test" profile (which also takes care of the database config and such). This is a first step for the whole profile story that might as well be merged right now.

TimvdLippe:
Yes I would prefer to do it completely in one go

PR #179, opened 2016-04-04
Token Code Optimilization

### LINKED ISSUE #164: Make one generic TokenRepository
Casting to correct token type can be done through a (currently non-existent) TokenService

### LINKED ISSUE #166: Make generic Token invalid exception
Currently the HttpRequest are made on the spot with hardcoded strings

Example:
https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java#L99

### LINKED ISSUE #168: Token UUID should be instantiated in the Token's constructor

bvanderdrift:
#164 In theory could be fixed further where there actually is one TokenRepository for all tokens, but Jpa doesn't support saving inherited Objects and be able to cast them back later.

Fixes #168 
Fixes #166 
Fixes #164


src/main/java/ch/wisv/areafiftylan/security/token/Token.java
```diff
@@ -54,7 +55,7 @@ public String getToken() {
         return token;
     }
 
-    public void setToken(String token) {
+    private void setToken(String token) {
```
skamoen: This is a pretty useless function if it's private. Might as well remove it

skamoen:
Great improvement, it all makes a lot of sense. The logic in the `AuthenticationController` might be better suited in a dedicated `TokenService`, but I'm not sure if that's already worth the trouble. Something to consider when Tokens are more widely used.

I'll merge after rebase and the quick fix of the comment I placed.

bvanderdrift:
Updated, back to @skamoen

skamoen:
LGTM

PR #180, opened 2016-04-04
User Cleanup

### LINKED ISSUE #97: User cleanup task
Just like expired Orders, unconfirmed Users should be deleted after some amount of minutes. This is a little more complicated then the Order due to various FK dependencies.

### LINKED ISSUE #152: Deal with expired verification tokens
Once the deadline for verification has passed, there's no way to verify an account anymore. There are two ways of solving this

1. Implement #97 to include unverified accounts with expired tokens
2. Allow for requesting another confirmation mail. 

I'm leaning towards 1.

bvanderdrift:
Fixes #152
Fixes #97

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/service/TaskScheduler.java
+++ b/src/main/java/ch/wisv/areafiftylan/service/TaskScheduler.java
@@ -1,12 +1,18 @@
 package ch.wisv.areafiftylan.service;
 
 import ch.wisv.areafiftylan.model.Order;
+import ch.wisv.areafiftylan.model.User;
 import ch.wisv.areafiftylan.model.util.OrderStatus;
+import ch.wisv.areafiftylan.security.token.VerificationToken;
 import ch.wisv.areafiftylan.service.repository.OrderRepository;
+import ch.wisv.areafiftylan.service.repository.UserRepository;
+import ch.wisv.areafiftylan.service.repository.token.VerificationTokenRepository;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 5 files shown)


src/main/java/ch/wisv/areafiftylan/service/TaskScheduler.java
```diff
@@ -44,8 +58,20 @@ public void ExpireOrders() {
...
+    }
+
     public static Predicate<Order> isExpired() {
         return o -> o.getStatus().equals(OrderStatus.CREATING) || o.getStatus().equals(OrderStatus.EXPIRED) ||
                 o.getStatus().equals(OrderStatus.CANCELLED);
     }
+
+    private void handleExpiredVerificationToken(VerificationToken verificationToken) {
+        verificationTokenRepository.delete(verificationToken);
+        userRepository.delete(verificationToken.getUser());
```
skamoen: This should be tested. There are quite a few database constraints which might prevent this from executing. 

This needs an integration test where a user is created through the controller, to account for possible future foreign key constraints

bvanderdrift: Didn't we already discuss this? This is untestable due to the time factor. Just like the expired orders functionality, I have tested this in postman which was succesful. Do you have any suggestions on how to test this?

skamoen: A simple manual invoke could test this function. This Scheduler can be autowired into the test, which could allow for this.

Also. There's no such thing as "untestable". Adten :stuck_out_tongue:

bvanderdrift:
Back to @skamoen

skamoen:
Yes, that's perfect. LGTM

PR #181, opened 2016-04-04
Add route to get all paid tickets for authenticated user

martijn9612:


src/main/java/ch/wisv/areafiftylan/controller/CurrentUserRestController.java
```diff
@@ -129,6 +124,19 @@
...
 
     /**
+     * Get the tickets of the currently logged in user. All the tickets owned by this user will be returned.
+     *
+     * @param auth The current User, injected by spring
+     *
+     * @return The current owned tickets, if any exist
+     */
+    @RequestMapping(value = "/tickets", method = RequestMethod.GET)
+    public Collection<Ticket> getAllTickets(Authentication auth) {
```
bvanderdrift: Why do we use List and Collection as return types in the CurrentUserRestController? Maybe it's better if we just use one return type.

src/main/java/ch/wisv/areafiftylan/service/TicketService.java
```diff
@@ -13,6 +15,8 @@
 
     Integer getNumberSoldOfType(TicketType type);
 
+    Collection<Ticket> findValidTicketsByOwnerUsername(String username);
```
bvanderdrift: I personally would use a User object as parameter. I feel like it is better to keep data together as much as possible (whilst still relevant ofcourse). It could be the case that later you'd like to do a check in the service on another property of a user (like if they have a profile, no profile means no tickets so you don't have to perform a query.)

Only narrow down to one data point (the username in this case) when performing the query (so in the repository this type of parameter is ok).

I think this an overall better programming style because it gives more flexibility at no cost.

src/test/java/ch/wisv/areafiftylan/OrderRestIntegrationTest.java
```diff
@@ -924,6 +924,43 @@ public void testGetTicketAvailability(){
...
+        given().
+            filter(sessionFilter).
+            header(login.getCsrfHeader()).
+        when().
+            get("/users/current/tickets").
+        then().
+            body("object", hasSize(0));
+        //@formatter:on
+    }
 }
```
bvanderdrift: What about the test when there is no user logged in and the route is queried? Does the API give back the correct result? In the currentUserController I don't see any @PreAuthorize annotation. The team queries have this, but order queries also miss this. I would argue that this is incorrect and should be implemented in the order and ticket routes.

skamoen: You're partially right. See the GET method for the way to do this and why

```
// To prevent 403 errors on this endpoint, we manually handle unauthenticated users, instead of a
        // preauthorize tag.
        if (auth != null) {
            // Get the currently logged in user from the autowired Authentication object.
            UserDetails currentUser = (UserDetails) auth.getPrincipal();
            User user = userService.getUserByUsername(currentUser.getUsername()).get();
```

bvanderdrift:
Back to @martijn9612

martijn9612:
So this should remove the security concerns you have about this route, correct? The route now only works when there is a user authenticated, otherwise a forbidden status code will be returned.

src/main/java/ch/wisv/areafiftylan/service/TicketService.java
skamoen: The UserService is the only class with access to the Repository. I think it's better to leave all the User logic to the Service. The username is something in the Spring Security auth object and therefore doesn't require any logic done by the controller. 

Most Service functions work references(such asid's or usernames) to prevent User objects floating through the entire application. This is safer from a persistence perspective.

PR #183, opened 2016-04-08
Codecov.io, Take 2

skamoen:


skamoen:
I kind of like this, but it needs some tweaking in CodeCov settings to provide more useful information. Ignoring files will help with this. to be continued.

TimvdLippe:
I think a more detailed report will be provided once master has a coverage report. Just like https://github.com/mockito/mockito/pull/389#issuecomment-207978632

TimvdLippe:
Should be updated. If you could try a merge from master into this branch, hopefully the coverage report should include more information as it can now generate a diff between 2 diffs.

skamoen:
I don't think this is useful enough yet. Maybe next year when we'll actually use all the code that's in it.

PR #184, opened 2016-04-10
Fix hasAnyTicket method in CurrentUserService

skamoen:
The CurrentUserService checks if a User has a ticket. This method should also verify if this ticket is actually valid (paid for) before returning true. 

Also fixed minor formatting and made fields private.


TimvdLippe:
LGTM

TimvdLippe:
:+1:

PR #185, opened 2016-04-11
Team fixes

skamoen:
This PR tweaks the Team functionality a bit to be more compatible with the front-end. A new Team View has been introduced, which displays additional information to Team members only. 

It also fixes a severe dataleak where Captains would get full member information on inviting a new member (very minor detail, really!).

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
@@ -190,7 +190,7 @@ public class TeamRestController {
     @PreAuthorize("@currentUserServiceImpl.canRevokeInvite(principal, #token)")
     @RequestMapping(method = RequestMethod.DELETE, value = "/invites")
     public ResponseEntity<?> declineTeamInvite(@RequestBody String token) {
-        teamService.removeInvite(token);
+        teamService.revokeInvite(token);
         return createResponseEntity(HttpStatus.OK, "Invite successfully declined");
     }
 
--- a/src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
@@ -90,7 +90,7 @@ public class TeamRestController {
      * @return A representation the Team with given Id.
      */
     @PreAuthorize("@currentUserServiceImpl.canAccessTeam(principal, #teamId)")
-    @JsonView(View.Public.class)
+    @JsonView(View.Team.class)
     @RequestMapping(method = RequestMethod.GET, value = "/{teamId}")
     public Team getTeamById(@PathVariable Long teamId) {
         return this.teamService.getTeamById(teamId);
@@ -104,7 +104,7 @@ public class TeamRestController {
      * @return A list of the members of the Team with the given Id. Public information only.
      */
     @PreAuthorize("@currentUserServiceImpl.canAccessTeam(principal, #teamId)")
-    @JsonView(View.Public.class)
+    @JsonView(View.Team.class)
     @RequestMapping(method = RequestMethod.GET, value = "/{teamId}/members")
     public Set<User> getTeamMembersById(@PathVariable Long teamId) {
         Team team = teamService.getTeamById(teamId);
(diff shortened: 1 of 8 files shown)


src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
```diff
@@ -156,9 +155,8 @@ public Team getTeamById(@PathVariable Long teamId) {
             "and @currentUserServiceImpl.hasAnyTicket(#username)")
```
bvanderdrift: Does a user that is to be invited need to have a ticket? I know this isn't implemented in this PR but I haven't seen this before. I thought the idea was that you could form a team, and then buy tickets for your team.

src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
```diff
@@ -156,9 +155,8 @@ public Team getTeamById(@PathVariable Long teamId) {
...
     @RequestMapping(method = RequestMethod.POST, value = "/{teamId}/invites")
     public ResponseEntity<?> inviteTeamMember(@PathVariable Long teamId, @RequestBody String username) {
-        TeamInviteToken teamInviteToken = teamService.inviteMember(teamId, username);
-        return createResponseEntity(HttpStatus.OK, "User " + username + " successfully invited to Team " + teamId,
-                teamInviteToken);
+        teamService.inviteMember(teamId, username);
+        return createResponseEntity(HttpStatus.OK, "User " + username + " successfully invited to Team " + teamId);
```
bvanderdrift: Why doesn't this part return the token anymore? Doesn't the frontend need the token to, for example, cancel an invite?

src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
```diff
@@ -167,17 +165,31 @@ public Team getTeamById(@PathVariable Long teamId) {
...
+    /**
+     * Decline or revoke an invite
+     *
+     * @param token Token of the Invite
+     *
+     * @return Statusmessage
+     */
     @PreAuthorize("@currentUserServiceImpl.canRevokeInvite(principal, #token)")
     @RequestMapping(method = RequestMethod.DELETE, value = "/invites")
     public ResponseEntity<?> declineTeamInvite(@RequestBody String token) {
```
bvanderdrift: Revoke sounds as if the sender can also cancel the invite. I wouldn't name this method 'declineTeam...' but 'revokeTeam...'

src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
skamoen: Yes, any Team related things require the user to have a Ticket. You can add users after you've sent them the ticket.

src/main/java/ch/wisv/areafiftylan/controller/TeamRestController.java
skamoen: It returned the token itself, which contains a lot more information then you want to return. The Front-end requests the tokens when it needs them. That uses the `TokenResponse` object, which only contains the information required.

src/main/java/ch/wisv/areafiftylan/service/TeamServiceImpl.java
```diff
@@ -89,8 +88,10 @@ public Team update(Long teamId, TeamDTO input) {
         }
 
         // If the Captain username is set, change the Captain
-        if (!Strings.isNullOrEmpty(input.getCaptainUsername())) {
-            User captain = userService.getUserByUsername(input.getCaptainUsername()).get();
+        String captainUsername = input.getCaptainUsername();
+        if (!Strings.isNullOrEmpty(captainUsername)) {
```
bvanderdrift: I would also perform a check if the captain in input is that same as the current captain. I know this doesn't change anything in functionality but I think it's better code. Less execution.

src/main/java/ch/wisv/areafiftylan/service/TeamServiceImpl.java
```diff
@@ -109,12 +110,17 @@ public TeamInviteToken inviteMember(Long teamId, String username) {
...
 
+        // Check if the member isn't already part of the team
+        if (team.getMembers().contains(user)) {
+            throw new IllegalArgumentException("User is already a member of this team");
+        }
+
         // Check if the member isn't already invited
-        Optional<TeamInviteToken> optionalInvite = teamInviteTokenRepository.findByUserUsername(username).stream().
-                filter(token -> token.getTeam().equals(team)).findFirst();
+        boolean userAlreadyInvited = teamInviteTokenRepository.findByUserUsername(username).stream()
```
bvanderdrift: I would extract this logic (as well as some other logic in this method) in seperate methods. This entire method is 30 lines long. In my opinion way too long. This does not only provide better code readability but also code reusability.

bvanderdrift:
Back to @skamoen

skamoen:
Updated. Back to @bvanderdrift

bvanderdrift:
Looking good!

PR #189, opened 2016-04-17
Event creation and registration

### LINKED ISSUE #71: No double entries for tournament registration
Users must not be allowed to register twice for a tournament with different teams.

skamoen:
This PR adds the possibilty to create Events for which Teams can register. 

Fixes #71

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -1,6 +1,7 @@
 package ch.wisv.areafiftylan.controller;
 
 import ch.wisv.areafiftylan.dto.EventDTO;
+import ch.wisv.areafiftylan.exception.EventException;
 import ch.wisv.areafiftylan.model.Event;
 import ch.wisv.areafiftylan.model.Team;
 import ch.wisv.areafiftylan.model.view.View;
@@ -82,6 +83,20 @@ public class EventRestController {
     }
 
     /**
+     * Delete the event with the given ID
+     *
+     * @param eventId ID of the event to be deleted
+     *
+     * @return The deleted event
+     */
+    @PreAuthorize("hasRole('ADMIN')")
+    @RequestMapping(value = "/{eventId}", method = RequestMethod.DELETE)
+    public ResponseEntity<?> updateEvent(@PathVariable Long eventId) {
+        Event event = eventService.deleteEvent(eventId);
+        return createResponseEntity(HttpStatus.OK, "Event successfully deleted", event);
+    }
+
+    /**
      * Register a team for an event. Only for captains and admin
      *
      * @param teamId  Id of the Team to be registered
@@ -131,5 +146,10 @@ public class EventRestController {
     @ExceptionHandler(IllegalArgumentException.class)
     public ResponseEntity<?> handleIllegalArgumentException(IllegalArgumentException e) {
         return createResponseEntity(HttpStatus.CONFLICT, e.getMessage());
+    }
+
+    @ExceptionHandler(EventException.class)
+    public ResponseEntity<?> handleEventException(EventException e) {
+        return createResponseEntity(HttpStatus.BAD_REQUEST, e.getMessage());
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -7,6 +7,7 @@ import ch.wisv.areafiftylan.model.view.View;
 import ch.wisv.areafiftylan.service.EventService;
 import ch.wisv.areafiftylan.service.TeamService;
 import com.fasterxml.jackson.annotation.JsonView;
+import org.hibernate.exception.GenericJDBCException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
@@ -37,11 +38,11 @@ public class EventRestController {
     }
 
     /**
-     * Add a new Event
+     * Add a new WebEvent
      *
-     * @param eventDTO EventDTO containing all info about the Event. Validated for null fields
+     * @param eventDTO EventDTO containing all info about the WebEvent. Validated for null fields
      *
-     * @return Message containing the location of the created Event
+     * @return Message containing the location of the created WebEvent
      */
     @RequestMapping(method = RequestMethod.POST)
     @PreAuthorize("hasRole('ADMIN')")
@@ -111,5 +112,10 @@ public class EventRestController {
     @RequestMapping(value = "/{eventId}/teams", method = RequestMethod.GET)
     public Collection<Team> getRegisteredTeamsForEvent(@PathVariable Long eventId) {
         return eventService.getTeamsForEvent(eventId);
+    }
+
+    @ExceptionHandler(GenericJDBCException.class)
+    public ResponseEntity<?> handleIllegalArgumentException(GenericJDBCException e) {
+        return createResponseEntity(HttpStatus.CONFLICT, e.getMessage());
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -7,7 +7,6 @@ import ch.wisv.areafiftylan.model.view.View;
 import ch.wisv.areafiftylan.service.EventService;
 import ch.wisv.areafiftylan.service.TeamService;
 import com.fasterxml.jackson.annotation.JsonView;
-import org.hibernate.exception.GenericJDBCException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
@@ -114,8 +113,8 @@ public class EventRestController {
         return eventService.getTeamsForEvent(eventId);
     }
 
-    @ExceptionHandler(GenericJDBCException.class)
-    public ResponseEntity<?> handleIllegalArgumentException(GenericJDBCException e) {
+    @ExceptionHandler(IllegalArgumentException.class)
+    public ResponseEntity<?> handleIllegalArgumentException(IllegalArgumentException e) {
         return createResponseEntity(HttpStatus.CONFLICT, e.getMessage());
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -62,6 +62,7 @@ public class EventRestController {
      *
      * @return A collection of all current events
      */
+    @JsonView(View.Public.class)
     @RequestMapping(method = RequestMethod.GET)
     public Collection<Event> getAllEvents() {
         return eventService.getAllEvents();
@@ -104,8 +105,8 @@ public class EventRestController {
      *
      * @return Message containing the result of the registration
      */
-    @RequestMapping(value = "/{eventId}", method = RequestMethod.POST)
     @PreAuthorize("@currentUserServiceImpl.canEditTeam(principal, #teamId)")
+    @RequestMapping(value = "/{eventId}", method = RequestMethod.POST)
     public ResponseEntity<?> registerTeamForEvent(@RequestBody Long teamId, @PathVariable Long eventId) {
         Team team = teamService.getTeamById(teamId);
         eventService.registerTeam(eventId, team);
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -37,11 +37,11 @@ public class EventRestController {
     }
 
     /**
-     * Add a new WebEvent
+     * Add a new Event
      *
-     * @param eventDTO EventDTO containing all info about the WebEvent. Validated for null fields
+     * @param eventDTO EventDTO containing all info about the Event. Validated for null fields
      *
-     * @return Message containing the location of the created WebEvent
+     * @return Message containing the location of the created Event
      */
     @RequestMapping(method = RequestMethod.POST)
     @PreAuthorize("hasRole('ADMIN')")
@@ -64,6 +64,21 @@ public class EventRestController {
     @RequestMapping(method = RequestMethod.GET)
     public Collection<Event> getAllEvents() {
         return eventService.getAllEvents();
+    }
+
+    /**
+     * Update the event with the given eventId
+     *
+     * @param eventDTO DTO containing the fields that need to be updated. Can contain null fields
+     * @param eventId  Id of the event to be updated
+     *
+     * @return Message containing the updated event
+     */
+    @PreAuthorize("hasRole('ADMIN')")
+    @RequestMapping(value = "/{eventId}", method = RequestMethod.PUT)
+    public ResponseEntity<?> updateEvent(@RequestBody EventDTO eventDTO, @PathVariable Long eventId) {
+        Event event = eventService.updateEvent(eventId, eventDTO);
+        return createResponseEntity(HttpStatus.OK, "Event successfully updated", event);
     }
 
     /**
--- a/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
@@ -25,13 +25,13 @@ import static ch.wisv.areafiftylan.util.ResponseEntityBuilder.createResponseEnti
  */
 @RestController
 @RequestMapping(value = "/events")
-public class EventController {
+public class EventRestController {
 
     private EventService eventService;
     private TeamService teamService;
 
     @Autowired
-    public EventController(EventService eventService, TeamService teamService) {
+    public EventRestController(EventService eventService, TeamService teamService) {
         this.eventService = eventService;
         this.teamService = teamService;
     }
--- a/src/main/java/ch/wisv/areafiftylan/dto/EventDTO.java
+++ b/src/main/java/ch/wisv/areafiftylan/dto/EventDTO.java
@@ -1,4 +1,4 @@
-package ch.wisv.areafiftylan.model;
+package ch.wisv.areafiftylan.dto;
 
 import lombok.Getter;
 import lombok.Setter;
--- a/src/main/java/ch/wisv/areafiftylan/model/Event.java
+++ b/src/main/java/ch/wisv/areafiftylan/model/Event.java
@@ -1,6 +1,10 @@
 package ch.wisv.areafiftylan.model;
 
+import ch.wisv.areafiftylan.model.view.View;
+import com.fasterxml.jackson.annotation.JsonView;
+
 import javax.persistence.*;
+import java.util.HashSet;
 import java.util.Set;
 
 /**
@@ -9,19 +13,25 @@ import java.util.Set;
 @Entity
 public class Event {
 
+    @JsonView(View.Public.class)
     private String eventName;
 
+    @JsonView(View.Public.class)
     @Id
     @GeneratedValue
     private Long id;
 
+    @JsonView(View.Public.class)
     private int teamSize;
 
+    @JsonView(View.Public.class)
     private int teamLimit;
 
+    @JsonView(View.Public.class)
     @ElementCollection(targetClass = Team.class, fetch = FetchType.EAGER)
     private Set<Team> registeredTeams;
 
+    @JsonView(View.Public.class)
     private boolean openForRegistration = false;
 
     public Event() {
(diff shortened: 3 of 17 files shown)


bvanderdrift:
To start off, do I understand it correctly that it is only possible to participate in an event as a team? What about individual games as Hearthstone. Don't you think it's a bit devious to create a single-man team?

src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param teamId  Id of the Team to be removed from the Event
+     * @param eventId Id of the Event the Team should be removed from
+     *
+     * @return Statusmessage about the result of the removal
+     */
+    @RequestMapping(value = "/{eventId}/teams", method = RequestMethod.DELETE)
+    @PreAuthorize("@currentUserServiceImpl.canEditTeam(principal, #teamId)")
+    public ResponseEntity<?> removeTeamFromEvent(@RequestBody Long teamId, @PathVariable Long eventId) {
+        Team team = teamService.getTeamById(teamId);
```
bvanderdrift: This method actually adds a team, doesn't remove it.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public Event updateEvent(Long eventId, EventDTO eventDTO) {
+        Event event = getEventById(eventId);
+
+        if (!Strings.isNullOrEmpty(eventDTO.getName())) {
+            event.setEventName(eventDTO.getName());
+        }
+
+        if (eventDTO.getTeamLimit() != null) {
+            event.setTeamLimit(eventDTO.getTeamLimit());
```
bvanderdrift: What if, upon updating, more teams have already registered than the new limit. I personally think you shouldn't be able to lower the limit if the current registration count exceeds the new limit. The admin should first delete some teams in that case.

src/main/java/ch/wisv/areafiftylan/dto/EventDTO.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Created by Sille Kamoen on 27-3-16.
+ */
+public class EventDTO {
+
+    @Getter
+    @Setter
+    @NotEmpty
+    String name = "";
+
```
bvanderdrift: What about some more fields? Such as a location of an image for on the site? Maybe a description, sponsor, prices? We could even think of some more.

src/main/java/ch/wisv/areafiftylan/service/EventService.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public Event deleteEvent(Long eventId);
+
+    /**
+     * Enable the Event for registration or close it.
+     *
+     * @param eventId             Id of the Event to be locked/unlocked
+     * @param openForRegistration Boolean whether the event should be open for registration or not.
+     */
+    public void setOpenForRegistration(Long eventId, boolean openForRegistration);
```
bvanderdrift: There is currently no end-point to do this. I think before merging this PR this functionality should be implemented as well.

src/main/java/ch/wisv/areafiftylan/model/Event.java
```diff
@@ -0,0 +1,96 @@
...
+    public void setTeamLimit(int teamLimit) {
+        this.teamLimit = teamLimit;
+    }
+
+    public Set<Team> getRegisteredTeams() {
+        return registeredTeams;
+    }
+
+    public boolean isOpenForRegistration() {
+        return openForRegistration;
```
bvanderdrift: I think the registration should also be marked as "closed" if the teamLimit has been reached.

src/main/java/ch/wisv/areafiftylan/model/Event.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public Event() {
+        //JPA ONLY
+    }
+
+    public Event(String eventName, int teamSize, int teamLimit) {
+        this.eventName = eventName;
+        this.teamSize = teamSize;
+        this.teamLimit = teamLimit;
+        this.registeredTeams = new HashSet<>(teamLimit);
```
bvanderdrift: The size of this Hash does not get updated if the teamLimit value gets updated.

src/main/java/ch/wisv/areafiftylan/service/EventService.java
```diff
@@ -0,0 +1,91 @@
...
+     * Add a new Event
+     *
+     * @param eventDTO DTO containing all information for the Event
+     *
+     * @return The newly created Event;
+     */
+    public Event addEvent(EventDTO eventDTO);
+
+    /**
+     * Change the Event with the given Id;
```
bvanderdrift: Personal autism, but I'd change this to "Update the..."

src/main/java/ch/wisv/areafiftylan/service/EventService.java
```diff
@@ -0,0 +1,91 @@
...
+    public Event updateEvent(Long eventId, EventDTO eventDTO);
+
+    /**
+     * Get all Teams currently registered for the Event
+     *
+     * @param eventId Id of the Event
+     *
+     * @return A Set of Teams currently registered for the Event
+     */
+    public Set<Team> getTeamsForEvent(Long eventId);
```
bvanderdrift: Why is a Set used here, and later on in this file a Collection?

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    private EventRepository eventRepository;
+
+    @Autowired
+    public EventServiceImpl(EventRepository eventRepository) {
+        this.eventRepository = eventRepository;
+    }
+
+    @Override
+    public void registerTeam(Long eventId, Team team) {
+        Event event = getEventById(eventId);
```
bvanderdrift: Why does this method require the Team and not teamId? I personally believe not logic should be in the Controller other than Authorization and responses.

src/main/java/ch/wisv/areafiftylan/controller/EventRestController.java
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param teamId  Id of the Team to be registered
+     * @param eventId Id of the Event to be registered for.
+     *
+     * @return Message containing the result of the registration
+     */
+    @PreAuthorize("@currentUserServiceImpl.canEditTeam(principal, #teamId)")
+    @RequestMapping(value = "/{eventId}", method = RequestMethod.POST)
+    public ResponseEntity<?> registerTeamForEvent(@RequestBody Long teamId, @PathVariable Long eventId) {
+        Team team = teamService.getTeamById(teamId);
```
bvanderdrift: I think the teamId should be passed on to the registerTeam method. Or the Event should be retrieved using the eventId. I prefer the first one.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        if (event.getTeamSize() != team.getSize()) {
+            throw new EventException("Team is not of the right size!");
+        }
+
+        if (event.getRegisteredTeams().size() > event.getTeamLimit()) {
```
bvanderdrift: Shouldn't this be >=?

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            throw new EventException("Event is full!");
+        }
+
+        if (!event.isOpenForRegistration()) {
+            throw new EventException("Event is not open for registration!");
+        }
+
+        // Check if one of the Team members is already a member
+        boolean duplicateUser = event.getRegisteredTeams().stream().flatMap(t -> t.getMembers().stream())
```
bvanderdrift: I'd rename this to hasDuplicateUser. duplicateUser insinuates that this is a User object.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+                .anyMatch(u -> team.getMembers().contains(u));
+
+        if(duplicateUser){
+            throw new EventException("A Member on your Team is already registered");
+        }
+
+        // We're registering the Team here
+        if (!event.addTeam(team)) {
+            throw new EventException("Something went wrong while registrating");
```
bvanderdrift: I think more feedback would be useful here. The Exception should be thrown by the addTeam method, and not return a bool which is later translated to a more generic exception.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        eventRepository.save(event);
+    }
+
+    @Override
+    public void removeTeamFromEvent(Long eventId, Team team) {
+        Event event = getEventById(eventId);
+        event.getRegisteredTeams().remove(team);
```
bvanderdrift: I think a checking if the team is actually a member couldn't hurt. This would provide more feedback to the frontend.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        eventRepository.save(event);
+    }
+
+    @Override
+    public Event addEvent(EventDTO eventDTO) {
+        eventRepository.findByEventName(eventDTO.getName()).ifPresent(e -> {
```
bvanderdrift: Maybe perform this checking case-insensitive?

src/main/java/ch/wisv/areafiftylan/service/repository/EventRepository.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import java.util.Optional;
+
+/**
+ * Created by Sille Kamoen on 27-3-16.
+ */
+public interface EventRepository extends JpaRepository<Event, Long> {
+
+    Optional<Event> findById(Long eventId);
```
bvanderdrift: The findOne method already does this. That doesn't use a wrapper though, but isn't checking a null return a good substitute?

bvanderdrift:
I have to head off now. I think there is some workable feedback in here. Haven't checked the api-frontend part and the tests yet.

src/main/java/ch/wisv/areafiftylan/dto/EventDTO.java
skamoen: This endpoint is not for the website, that's what the `/web` endpoints are for. There is some overlap in functionality, certainly when the intranet comes into play as well. This will probably move to the intranet alltogether for next year, with a dedicated `web` section that can be imported into the intranet. 2017 stuff.

src/main/java/ch/wisv/areafiftylan/model/Event.java
skamoen: That's alright. Hashsets are automatically resized when exceeding their size. This is just the initial allocation of memory. Not really needed, just slightly more optimized.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
skamoen: Throwing exceptions in model object is a bad idea.

skamoen:
Great feedback, thanks.

As far as the individual teams are concerned, it's a very small effort to create a personal team. It's a lot of effort and very error-prone to accept both Users and Teams for an Event. You'd have to work with different subclasses of Event, which doesn't sound desirable. I think it makes sense to stick with Teams on this one.

src/main/java/ch/wisv/areafiftylan/dto/EventDTO.java
bvanderdrift: Sounds good

src/main/java/ch/wisv/areafiftylan/model/Event.java
bvanderdrift: Allright. Might be confusing. What is the optimization? A simple google query does not return any pre-allocation optimization explanation.

src/main/java/ch/wisv/areafiftylan/service/EventServiceImpl.java
bvanderdrift: Why? And if this is true, I still think we should come up with a way to be able to provide more feedback upon failing instead of "Something went wrong."

bvanderdrift:
Good point. I've made this into an issue (#197).

src/main/java/ch/wisv/areafiftylan/web/service/WebEventsService.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.Collection;
+
+@Service
+public class WebEventsService {
+
+    public Collection<WebEvent> getAllEvents() {
+        return getDummyEvents();
```
bvanderdrift: Before the PR is completed this should be real data

src/main/java/ch/wisv/areafiftylan/web/service/WebSponsorServiceImpl.java
```diff
@@ -7,7 +7,7 @@
 import java.util.Collection;
 
 @Service
-public class SponsorServiceImpl {
+public class WebSponsorServiceImpl {
 
     public Collection<Sponsor> getAllSponsors() {
         return getDummySponspors();
```
bvanderdrift: Before the PR is completed this should be real data

src/main/java/ch/wisv/areafiftylan/web/service/WebTournamentServiceImpl.java
```diff
@@ -13,7 +13,7 @@
 import java.util.LinkedList;
 
 @Service
-public class TournamentServiceImpl {
+public class WebTournamentServiceImpl {
 
     public Collection<Tournament> getAllTournaments() {
         return getDummyTournaments();
```
bvanderdrift: Before the PR is completed this should be real data

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+    UserRepository userRepository;
+
+    @Before
+    public void setupEventTests() {
+        testEvent = new Event("Test Event", 2, 5);
+        testEvent.setOpenForRegistration(true);
+        testEvent = eventRepository.save(testEvent);
+
+        teamCaptain = userRepository
+                .save(new User("captain", new BCryptPasswordEncoder().encode("password"), "captain@team.com"));
```
bvanderdrift: Please use the new password methodology. We are using a single final string in the tests instead of hardcoded strings.

src/main/java/ch/wisv/areafiftylan/model/Event.java
```diff
@@ -0,0 +1,96 @@
...
+    private boolean openForRegistration = false;
+
+    public Event() {
+        //JPA ONLY
+    }
+
+    public Event(String eventName, int teamSize, int teamLimit) {
+        this.eventName = eventName;
+        this.teamSize = teamSize;
+        this.teamLimit = teamLimit;
```
bvanderdrift: Shoudl teamLimit < 2 be possible?

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+    public void cleanupEventTests() {
+        eventRepository.deleteAll();
+        teamRepository.deleteAll();
+    }
+
+    //region Test Add Events
+
+    @Test
+    public void testAddEventAsAnon() {
+        Map<String, String> eventDTO = new HashMap<>();
```
bvanderdrift: I'd extract this into a seperate method. It's a lot of duplicate code througout the tests, and using a method even gives some parameter control.

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+            content(eventDTO).contentType(ContentType.JSON).
+            post(EVENTS_ENDPOINT).
+        then().
+            statusCode(HttpStatus.SC_CREATED).
+            header("location", containsString("/events/")).
+            body("object.eventName", is("event"));
+        //@formatter:on
+    }
+
+    private void addEvent(String username, String name, String teamLimit, String teamSize) {
```
bvanderdrift: I'd group the Tests together as well as the private helper methods. This makes for an easier search in code.

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+            filter(sessionFilter).
+            header(login.getCsrfHeader()).
+        when().
+            content(eventDTO).contentType(ContentType.JSON).
+            post(EVENTS_ENDPOINT).
+        then().
+            statusCode(HttpStatus.SC_CREATED).
+            header("location", containsString("/events/")).
+            body("object.eventName", is("event"));
+        //@formatter:on
```
bvanderdrift: I'd also logout in this method. This makes sure the global state stays the same upon calling this code (except for the known addition of an event.) Else, this should be named addEventAndLogin.

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
bvanderdrift: Why not return the response? This makes for easier testing. In some tests you do this manually and check the response. This is also done in other tests (TicketTransfer)

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+    @Test
+    public void testAddEventDuplicateNameAsAdmin() {
+        Map<String, String> eventDTO = new HashMap<>();
+        eventDTO.put("name", "event");
+        eventDTO.put("teamLimit", "10");
+        eventDTO.put("teamSize", "2");
+
+        addEvent("admin", eventDTO.get("name"), eventDTO.get("teamLimit"), eventDTO.get("teamSize"));
+
+        SessionData login = login("admin");
```
bvanderdrift: Why not use admin.getUsername() for more consistent code?

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+            filter(sessionFilter).
+            header(login.getCsrfHeader()).
+        when().
+            content(eventDTO).contentType(ContentType.JSON).
+            post(EVENTS_ENDPOINT).
+        then().
+            statusCode(HttpStatus.SC_CONFLICT);
+        //@formatter:on
+
+        Assert.assertEquals(2, eventRepository.count());
```
bvanderdrift: Wouldn't you expect a single event here? Do I miss something?

bvanderdrift: Nevermind, I now see that in the @Before a new event is made.

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+        when().
+            content(eventDTO).contentType(ContentType.JSON).
+            put(EVENTS_ENDPOINT + "/" + testEvent.getId()).
+        then().
+            statusCode(HttpStatus.SC_FORBIDDEN);
+        //@formatter:on
+
+        Event changedEvent = eventRepository.findOne(testEvent.getId());
+
+        Assert.assertEquals(testEvent.getEventName(), changedEvent.getEventName());
```
bvanderdrift: teamSize is also changed. I know this would be a weird case, but I'd check the equality of this property as well.

src/test/java/ch/wisv/areafiftylan/EventRestIntegrationTest.java
```diff
@@ -0,0 +1,769 @@
...
+        //@formatter:off
+        given().
+            filter(sessionFilter).
+            header(login.getCsrfHeader()).
+        when().
+            content(eventDTO).contentType(ContentType.JSON).
+            put(EVENTS_ENDPOINT + "/999").
+        then().
+            statusCode(HttpStatus.SC_BAD_REQUEST);
+        //@formatter:on
```
bvanderdrift: Also weird case, but maybe check the original event to be sure the data hasn't actually changed. Returning a bad request is not a assurance that the internal hasn't actually changed.

mgajanssen:
I assume this entire feature is going to get removed due to the fact that it is redundant, how harsh this may be? This is/will be replaced by the echo-gaming project or another solution. Is that correct? Could you shed some light onto this @skamoen ?

skamoen:
No plans for this yet. Should be discussed with the new LANcie.

mgajanssen:
Sure, sounds good.

skamoen:
Closing this for now, but let's keep the branch for reference.

PR #193, opened 2016-04-22
Throw error on duplicate ticket transfer setup

martijn9612:
This functionality fixes the issue with requesting multiple ticket transfer setups for the same ticket.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
@@ -1,5 +1,6 @@
 package ch.wisv.areafiftylan.controller;
 
+import ch.wisv.areafiftylan.exception.DuplicateTicketTransferTokenException;
 import ch.wisv.areafiftylan.security.token.TicketTransferToken;
 import ch.wisv.areafiftylan.service.TicketService;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -41,5 +42,10 @@ public class TicketTransferRestController {
         ticketService.cancelTicketTransfer(token);
 
         return createResponseEntity(HttpStatus.OK, "Ticket transfer successfully cancelled");
+    }
+
+    @ExceptionHandler(DuplicateTicketTransferTokenException.class)
+    public ResponseEntity<?> handleDuplicateTicketTransFerException(DuplicateTicketTransferTokenException ex) {
+        return createResponseEntity(HttpStatus.BAD_REQUEST, ex.getMessage());
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
@@ -1,13 +1,19 @@
 package ch.wisv.areafiftylan.controller;
 
 import ch.wisv.areafiftylan.exception.DuplicateTicketTransferTokenException;
+import ch.wisv.areafiftylan.model.view.View;
 import ch.wisv.areafiftylan.security.token.TicketTransferToken;
 import ch.wisv.areafiftylan.service.TicketService;
+import com.fasterxml.jackson.annotation.JsonView;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.web.bind.annotation.*;
+
+import java.util.Collection;
 
 import static ch.wisv.areafiftylan.util.ResponseEntityBuilder.createResponseEntity;
 
@@ -42,6 +48,15 @@ public class TicketTransferRestController {
         ticketService.cancelTicketTransfer(token);
 
         return createResponseEntity(HttpStatus.OK, "Ticket transfer successfully cancelled");
+    }
+
+    @PreAuthorize("isAuthenticated() and @currentUserServiceImpl.hasAnyTicket(principal)")
+    @RequestMapping(value = "/tickets/tokens", method = RequestMethod.GET)
+    public ResponseEntity<?> getTicketTokensOpenForTransfer(Authentication auth) {
+        UserDetails currentUser = (UserDetails) auth.getPrincipal();
+        Collection<TicketTransferToken> tokens = ticketService.getTicketTransferTokensByUser(currentUser.getUsername());
+
+        return createResponseEntity(HttpStatus.OK, "Ticket transfer tokens successfully retrieved", tokens);
     }
 
     @ExceptionHandler(DuplicateTicketTransferTokenException.class)
--- a/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
@@ -1,10 +1,8 @@
 package ch.wisv.areafiftylan.controller;
 
 import ch.wisv.areafiftylan.exception.DuplicateTicketTransferTokenException;
-import ch.wisv.areafiftylan.model.view.View;
 import ch.wisv.areafiftylan.security.token.TicketTransferToken;
 import ch.wisv.areafiftylan.service.TicketService;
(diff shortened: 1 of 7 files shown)


src/main/java/ch/wisv/areafiftylan/service/TicketServiceImpl.java
```diff
@@ -84,18 +83,22 @@ public TicketTransferToken setupForTransfer(Long ticketId, String goalUserName){
         User u = userService.getUserByUsername(goalUserName).orElseThrow(() -> new UsernameNotFoundException("User " + goalUserName + " not found."));
         Ticket t = ticketRepository.findOne(ticketId);
 
-        TicketTransferToken ttt = new TicketTransferToken(u, t);
+        if (tttRepository.findByTicketId(ticketId).isPresent()) {
```
bvanderdrift: Damn, as I currently see it the tttRepository can only query for one token. It's possible that there are multiple. This has to be fixed. Also this currently only checks the presence of such a token, not if it's valid.

src/main/java/ch/wisv/areafiftylan/service/repository/token/TicketTransferTokenRepository.java
```diff
@@ -8,4 +8,6 @@
 
 @Repository
 public interface TicketTransferTokenRepository extends TokenRepository<TicketTransferToken> {
+
+    Optional<TicketTransferToken> findByTicketId(Long ticketId);
```
bvanderdrift: Aah it seems you have added this Query. This should be findAllByTicketId.

src/test/java/ch/wisv/areafiftylan/TicketTransferRestIntegrationTest.java
```diff
@@ -199,6 +199,31 @@ public void testDoTransfer_Outsider(){
     }
 
     @Test
+    public void testDuplicateSetupForTransfer() {
+        SessionData login = login(user.getUsername(), userCleartextPassword);
+
+        //@formatter:off
+        given().
```
bvanderdrift: Please use the addTicketTransfer method. This is duplicate code.

martijn9612:
All your remarks and concerns should now be fixed, back to you @bvanderdrift!

src/test/java/ch/wisv/areafiftylan/TicketTransferRestIntegrationTest.java
```diff
@@ -199,6 +199,24 @@ public void testDoTransfer_Outsider(){
     }
 
     @Test
+    public void testDuplicateSetupForTransfer() {
+        addTicketTransfer(user.getUsername(), userCleartextPassword);
+
+        SessionData login = login(user.getUsername(), userCleartextPassword);
+
+        //@formatter:off
+        given()
```
bvanderdrift: This can also be replaced by addTicketTransfer. That method returns the response which can then be checked.

src/main/java/ch/wisv/areafiftylan/service/TicketServiceImpl.java
```diff
@@ -84,18 +85,26 @@ public TicketTransferToken setupForTransfer(Long ticketId, String goalUserName){
...
         Ticket t = ticketRepository.findOne(ticketId);
 
-        TicketTransferToken ttt = new TicketTransferToken(u, t);
+        List<TicketTransferToken> ticketTransferTokens = tttRepository.findAllByTicketId(ticketId).stream()
+                .filter(Token::isValid)
+                .collect(Collectors.toList());
 
-        tttRepository.save(ttt);
+        if (!ticketTransferTokens.isEmpty()) {
+            throw new IllegalStateException("Ticket " + ticketId + " is already set up for transfer!");
```
bvanderdrift: I would think about making a special exception for this. This is more in line with our programming style and makes sure the exceptions (with the message strings) are nicely bundled in the exceptions folder instead of scrambled in all the code.

bvanderdrift:
Minor comments, but other than that it looks good!

src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
```diff
@@ -42,4 +43,9 @@ public TicketTransferRestController(TicketService ticketService) {
 
         return createResponseEntity(HttpStatus.OK, "Ticket transfer successfully cancelled");
     }
+
+    @ExceptionHandler(DuplicateTicketTransferTokenException.class)
+    public ResponseEntity<?> handleDuplicateTicketTransFerException(DuplicateTicketTransferTokenException ex) {
```
bvanderdrift: Little styling issue : "...TicketTrans**Fer**Exc..."

src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
```diff
@@ -44,6 +50,15 @@ public TicketTransferRestController(TicketService ticketService) {
         return createResponseEntity(HttpStatus.OK, "Ticket transfer successfully cancelled");
     }
 
+    @PreAuthorize("isAuthenticated() and @currentUserServiceImpl.hasAnyTicket(principal)")
```
bvanderdrift: Why is the "hasAnyTicket" a security/authorization check? A user may still query right, but would just get an empty set as a result.

src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
```diff
@@ -44,6 +50,15 @@ public TicketTransferRestController(TicketService ticketService) {
...
+    @RequestMapping(value = "/tickets/tokens", method = RequestMethod.GET)
+    public ResponseEntity<?> getTicketTokensOpenForTransfer(Authentication auth) {
+        UserDetails currentUser = (UserDetails) auth.getPrincipal();
+        Collection<TicketTransferToken> tokens = ticketService.getTicketTransferTokensByUser(currentUser.getUsername());
+
+        return createResponseEntity(HttpStatus.OK, "Ticket transfer tokens successfully retrieved", tokens);
```
bvanderdrift: I'd like to see a '.' at the end of a sentence.

src/main/java/ch/wisv/areafiftylan/service/TicketServiceImpl.java
```diff
@@ -132,6 +132,13 @@ public void cancelTicketTransfer(String token){
         tttRepository.save(ttt);
     }
 
+    @Override
+    public Collection<TicketTransferToken> getTicketTransferTokensByUser(String username) {
```
bvanderdrift: You're not performing a query for all TicketTransferTokens but only the valid ones. Please rename the method to represent the query.

bvanderdrift: I'd also like to add that this query might have some problems. If the ticket is transferred the TicketOwnerUsername changes to the receiver. In this case the TicketOwnerUsername and the Token User would be the same. 

Short version: the underlying data changes once the transfer happens, and so the result of this query changes.

src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
bvanderdrift: I don't know why GH marks this as outdated. It isn't. I'll re-comment for visibility.

src/main/java/ch/wisv/areafiftylan/controller/TicketTransferRestController.java
```diff
@@ -42,4 +49,18 @@ public TicketTransferRestController(TicketService ticketService) {
 
         return createResponseEntity(HttpStatus.OK, "Ticket transfer successfully cancelled");
     }
+
+    @PreAuthorize("isAuthenticated() and @currentUserServiceImpl.hasAnyTicket(principal)")
```
bvanderdrift: Why is the "hasAnyTicket" a security/authorization check? A user may still query right, but would just get an empty set as a result.

src/main/java/ch/wisv/areafiftylan/service/TicketServiceImpl.java
martijn9612: This is not unexpected behavior. Once the ticket is transferred, the user no longer has any control over this ticket. The user will not see the ticket because the TicketTransferToken is gone, also he is not able to set up a transfer. The fact that the TicketOwnerUsername changes is expected and correct behavior.

bvanderdrift:
LGTM. Merging.

PR #201, opened 2016-05-02
Clear previous Seat on change

skamoen:
When the User changes his Seat, the previously reserved seat attached to the ticket should be cleared. This PR fixes that and updates the test to verify.

Also some supersmall fixes and formatter pass.


TimvdLippe:
LGTM

PR #203, opened 2016-05-02
Removed the Optional<Ticket> method from the Ticket Repository

skamoen:
User can have multple Tickets, this will throw an exception when that is the case.


TimvdLippe:
Could you add a regression test to prevent future problems?

skamoen:
I removed the method. Does that count?

src/main/java/ch/wisv/areafiftylan/service/CurrentUserServiceImpl.java
```diff
@@ -148,8 +147,9 @@ public boolean hasAnyTicket(Object principal) {
 
     @Override
     public boolean hasAnyTicket(String username) {
-        Optional<Ticket> ticket = ticketRepository.findByOwnerUsername(username);
-        return ticket.isPresent() && ticket.get().isValid();
+        Collection<Ticket> tickets = ticketRepository.findAllByOwnerUsername(username);
```
TimvdLippe: Then I would turn this in a one-liner.

TimvdLippe:
LGTM

PR #205, opened 2016-05-05
Fix check for duplicate teams with other casing

### LINKED ISSUE #204: Teamnames are checked case-sensitive for duplicity
I.e. "Lancie" and "lancie" can both be made.

bvanderdrift:
Fixes #204


martijn9612:
LGTM

PR #208, opened 2016-05-07
Token based authentication

skamoen:
For clients that are not the browser, the form-based authentication doesn't work. This implements a Token-based authentication method that generates a Token with which clients can perform actions.

Tokens are generated with a `POST` on `/token` with a `username` and `password`. If credentials are correct, a Token will be returned and saved.

Requests with a token in the `X-Auth-Token` header will be authenticated by the new Filter. 

Requests with a valid token are excluded from CSRF.


src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
```diff
@@ -63,19 +73,43 @@ protected void configure(HttpSecurity http) throws Exception {
...
+    // We want to enable CSRF for State Changing methods. We don't want CSRF for requests with valid tokens.
+    private RequestMatcher csrfRequestMatcher = new RequestMatcher() {
+        private Pattern allowedMethods = Pattern.compile("^(GET|HEAD|TRACE|OPTIONS)$");
+
+        @Override
+        public boolean matches(HttpServletRequest request) {
+            String requestHeader = request.getHeader("X-Auth-Token");
+            boolean methodAllowed = !(allowedMethods.matcher(request.getMethod()).matches());
+            boolean validAuthToken =
+                    !authenticationTokenRepository.findByToken(requestHeader).filter(Token::isValid).isPresent();
```
praseodym: The `TokenAuthenticationFilter` will also check this later in the chain so this is a little redundant.

src/main/java/ch/wisv/areafiftylan/security/TokenAuthenticationFilter.java
```diff
@@ -0,0 +1,48 @@
...
+        if (!Strings.isNullOrEmpty(xAuth)) {
+            AuthenticationToken authenticationToken = authenticationTokenRepository.findByToken(xAuth)
+                    .orElseThrow(() -> new TokenNotFoundException(xAuth));
+
+            if (authenticationToken.isValid()) {
+                User user = authenticationToken.getUser();
+                SecurityContextHolder.getContext()
+                        .setAuthentication(new PreAuthenticatedAuthenticationToken(user, "N/A", user.getAuthorities()));
+            } else {
+                throw new NonceExpiredException("Token expired!");
```
praseodym: Nonce?

src/main/java/ch/wisv/areafiftylan/security/token/AuthenticationToken.java
```diff
@@ -0,0 +1,25 @@
...
+    public AuthenticationToken() {
+        // JPA ONLY
+    }
+
+    public AuthenticationToken(User user, int expiration) {
+        super(user, expiration);
+    }
+
+    public PreAuthenticatedAuthenticationToken getAuthenticationToken() {
+        return new PreAuthenticatedAuthenticationToken(this.getUser(), "N/A");
```
praseodym: Not used

src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
skamoen: This is here because I don't want to disable CSRF for any request with a random "X-Auth-Token" header. How would you do this in a less redundant way?

skamoen:
This is not needed on master. Will use for development of the intranet and create a new PR when that's going live.

PR #209, opened 2016-05-07
Flyway migration

### LINKED ISSUE #207: Use Liquibase for database migration
Now that we have significant persistence changes, we need to deal with them

http://www.baeldung.com/liquibase-refactor-schema-of-java-app

Section 6.2 sounds like the thing we want.

skamoen:
This automates schema updates when we change our models. 

Works (twice) on my machine, but should be tested by the reviewer to make sure it doesn't kill the database.

Don't forget to gradle refresh and update properties

Fixes #207


skamoen:
Todo when migrating:
- [ ] Remove locked_for_transfer column from Ticket table

martijn9612:
All works well when the `locked_for_transfer` column is deleted. Merging this.

PR #214, opened 2016-05-09
Make all String-related operations case-insensitive

skamoen:
Case-sensitive stuff is annoying and is the most heard complaint. Let's fix this. 

I updated all repositories, but I haven't had the time to properly consider all aspects. If everyone can add a testcase or two to every integrationtest, we can deploy this ASAP

@martijn9612 @bvanderdrift

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
@@ -32,7 +32,7 @@ public class SeatServiceImpl implements SeatService {
 
     @Override
     public List<Seat> getSeatsByUsername(String username) {
-        return seatRepository.findByTicketOwnerUsername(username);
+        return seatRepository.findByTicketOwnerUsernameIgnoreCase(username);
 
     }
 
@@ -47,7 +47,7 @@ public class SeatServiceImpl implements SeatService {
 
     @Override
     public SeatmapResponse getSeatGroupByName(String groupname) {
-        List<Seat> seatGroup = seatRepository.findBySeatGroup(groupname);
+        List<Seat> seatGroup = seatRepository.findBySeatGroupIgnoreCase(groupname);
 
         Map<String, List<Seat>> seatMapResponse = new HashMap<>();
         seatMapResponse.put(groupname, seatGroup);
@@ -63,7 +63,7 @@ public class SeatServiceImpl implements SeatService {
 
         // We can only reserve one seat at a time, to prevent the extremely unlikely event of simultaneous requests
         synchronized (seatReservationLock) {
-            Seat seat = seatRepository.findBySeatGroupAndSeatNumber(groupname, seatnumber);
+            Seat seat = seatRepository.findBySeatGroupIgnoreCaseAndSeatNumber(groupname, seatnumber);
             previousSeat.ifPresent(s -> s.setTicket(null));
             if (!seat.isTaken()) {
                 seat.setTicket(ticket);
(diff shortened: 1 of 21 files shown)


skamoen:
I added case-related tests for every integration test. Required some small fixes. Ready for review!

martijn9612:
Looks good, is there a reason why you changed the indentation in the `.when()` and `.then()` statements?

skamoen:
That's the autoformatter at work without `//@formatter:off`. Doesn't really matter.

src/main/java/ch/wisv/areafiftylan/service/SeatServiceImpl.java
```diff
@@ -47,10 +47,10 @@ public SeatmapResponse getAllSeats() {
 
     @Override
     public SeatmapResponse getSeatGroupByName(String groupname) {
-        List<Seat> seatGroup = seatRepository.findBySeatGroup(groupname);
+        List<Seat> seatGroup = seatRepository.findBySeatGroupIgnoreCase(groupname);
 
         Map<String, List<Seat>> seatMapResponse = new HashMap<>();
-        seatMapResponse.put(groupname, seatGroup);
+        seatMapResponse.put(seatGroup.get(0).getSeatGroup(), seatGroup);
```
bvanderdrift: Border-case seatGroup is empty throws an ArrayOutOfBounds error here. Even though I'm sure this is not likely to happen I think it's better to make sure no undefined errors can be thrown.

src/main/java/ch/wisv/areafiftylan/service/UserServiceImpl.java
```diff
@@ -68,7 +69,9 @@ public User getUserById(long id) {
     }
 
     @Override
-    public User create(UserDTO userDTO, HttpServletRequest request) {
+    public User create(UserDTO userDTO, HttpServletRequest request) throws DataIntegrityViolationException {
+        checkIfFieldsInUse(userDTO);
```
bvanderdrift: A check should return a boolean value and not 'blindly' throw an Exception in my eyes. The responsibility is a check, not to handle that check. If that would be the case the function could be renamed for better code readability to "checkIfFieldInUseAndHandleResult" or something which is better. I would still prefer to return a boolean or maybe even an enum that returns what is in use: "None", "Username" or "Email".

src/main/java/ch/wisv/areafiftylan/service/UserServiceImpl.java
```diff
@@ -243,18 +252,7 @@ private String getPasswordHash(String plainTextPassword) {
...
-    /**
-     * Turns the request into a url to which the request is made. For example https://localhost:8080 TODO: This makes a
-     * call directly to the API, this should be handled by the front-end instead
-     *
-     * @param request The HttpServletRequest of the call that is made
-     *
-     * @return Formatted string of the base URL
-     */
-    public String getAppUrl(HttpServletRequest request) {
-        return request.getScheme() + "://" + request.getServerName() + ":" + request.getServerPort();
```
bvanderdrift: Good find

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -98,32 +98,41 @@ public void testAddTransfer_Owner(){
...
+        Assert.assertTrue(tttRepository.count() == 0);
+    }
+
+    @Test
+    public void testDoTransfer_Anon() {
         TicketTransferToken ttt = addTicketTransferGetToken();
 
         given().
-        when().
+                when().
```
bvanderdrift: Just curious, but why all these introduced tabs? Is that due to copy-pasting or is there a specific reason>?

martijn9612: https://github.com/AreaFiftyLAN/lancie-api/pull/214#issuecomment-217851962
I think it is because the auto formatter is off.

bvanderdrift:
Can't think of other testcases right now. Next to the minor comments all looks good to me!

skamoen:
I removed support for SeatGroup related case-insensitive search as it doesn't make that much sense and the border-case you mention requires more code then it's worth. 

Renamed the method to better represent what it does. Dealing with private ENUM's, and handling them afterwards seems devious compared to the simple method it currently is. 

Back to @bvanderdrift

martijn9612:
Looks good.

PR #215, opened 2016-05-09
Set up logging for the api

martijn9612:
This adds logging functionality with separated log files for different log levels. Logging for password reset is added due to necessity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ node_modules
 dist
 bower_components
 .tmp
+
+# log folder
+logs/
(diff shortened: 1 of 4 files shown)


bvanderdrift:
LGTM. Have to admit I do not have full knowledge of the logging system. Could you create an issue and reference this next time, or maybe do it even after the merge for this issue? This gives us clearer insight in why the password reset logging is a necessity.

martijn9612:
We have an issue for that some users do not receive an email when requesting a password reset. By logging this, we hope to find the problem. There was no issue created for this, although there should have been.

PR #225, opened 2016-05-18
TicketType deadlines

skamoen:
This PR adds deadlines to TicketType. With this, you solve two usecases. 
1. Once the event started, you don't want to sell tickets anymore.
2. To set deadlines, you want to introduce time-sensitive Ticket Types

For the last case, the Last Minute ticket was added to the enum. Hard to test, as this depends on a hardcoded enum and LocalDateTime.now(). Open for suggestions, but as these are hardcoded values in the Enum with trivial comparisons, it _should_ be fine.

TODO:
- [x] Determine deadline
- [x] Deteremine global ticket limit


bvanderdrift:
Deadline: 28 of May 23:59 for clarity. Other than that LGTM

PR #230, opened 2016-05-20
Add route to get all tickets

martijn9612:


skamoen:
This looks trivial enough to me

PR #231, opened 2016-05-20
Add route to get all tickets with pickup

martijn9612:


src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -429,4 +430,48 @@ private User createTeamReturnSingleMember() {
...
+        .then()
+            .statusCode(HttpStatus.SC_FORBIDDEN);
+        //@formatter:on
+    }
+
+    @Test
+    public void testGetAllTicketsForTransport_User() {
+        SessionData login = login("user", userCleartextPassword);
+
+
```
skamoen: Useless white line. Also, use `user.getUsername()` here to make it consistent.

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -429,4 +430,48 @@ private User createTeamReturnSingleMember() {
...
+            .header(login.getCsrfHeader())
+        .when()
+            .get("/tickets/transport")
+        .then()
+            .statusCode(HttpStatus.SC_FORBIDDEN);
+        //@formatter:on
+    }
+
+    @Test
+    public void testGetAllTicketsForTransport_Admin() {
```
skamoen: This test is not sufficient by itself. It needs at least another test where there is a difference between all tickets and transport tickets.

skamoen:
Some small remarks. Also needs a rebase, other then that looks good.

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
martijn9612: Ah, that is indeed a better thing to do!

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
martijn9612: The `ticketRepository` is initialized with a ticket without transport. This can be seen in the `@before` method. The `ticketRepository` contains 2 tickets when executing this test. That should be enough, or am I missing something?

skamoen: Maybe add a check that you get the ticket with `pickupSerivce = true` then?

martijn9612: My bad, this is not happening right now. I'll fix it!

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -471,4 +472,49 @@ private User createTeamReturnSingleMember() {
...
+
+        //@formatter:off
+        given()
+            .filter(sessionFilter)
+            .header(login.getCsrfHeader())
+        .when()
+            .get("/tickets/transport")
+        .then()
+            .statusCode(HttpStatus.SC_OK)
+            .body("", hasSize(1));
```
skamoen: I'm not seeing the extra check mentioned in the commit message?

skamoen:
LGTM

PR #236, opened 2016-05-21
Fix: Invalid Ticket Seat reservation

skamoen:
Fixed a User being able to reserve a seat with an invalid (unpaid) ticket.

### DIFF
--- a/src/test/java/ch/wisv/areafiftylan/SeatRestIntegrationTest.java
+++ b/src/test/java/ch/wisv/areafiftylan/SeatRestIntegrationTest.java
@@ -504,6 +504,25 @@ public class SeatRestIntegrationTest extends IntegrationTest {
     }
 
     @Test
+    public void reserveSeatWithInvalidTicket() {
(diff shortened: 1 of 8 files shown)


src/test/java/ch/wisv/areafiftylan/SeatRestIntegrationTest.java
```diff
@@ -504,6 +504,25 @@ public void reserveTakenSeatAsUser() {
     }
 
     @Test
+    public void reserveSeatWithInvalidTicket() {
+        userTicket.setValid(false);
+        ticketRepository.save(userTicket);
+
+        SessionData login = login("user", userCleartextPassword);
```
martijn9612: Please use the username from the user variable.

martijn9612:
Changes look good, only a build fix is required. This was a really sneaky way to bypass paying for your ticket!

martijn9612:
LGTM!

PR #241, opened 2016-05-23
Consumption Tracking

### LINKED ISSUE #211: Consumption Tracking
### Consumption Tracking
Users should be able to consume things during the event. This will contain, but is not limited to, dinner, breakfast and goodies they are entitled to.

##### Requirements
* There is a dynamic master set of consumptions.
* Every user should have a set of consumption values based on the master set.
* It should be possible to add/remove consumption types while values has already been altered for users.
* There should be an endpoint to retrieve a user's consumption values.
* There should be an endpoint to set a user's consumption value.

bvanderdrift:
Fixes #211 

## Consumption Tracking
This PR contains the feature for consumption tracking. Only Admins have access to this endpoint, since during the event admins need to verify if a user has made a consumption and alter the consumption state. Consumptions are identified by an id. 

#### Endpoints

* ```GET "/consumptions"```: returns if a consumption, defined by a ```ConsumptionDTO```, has been done 
* ```GET "/consumptions/{ticketId}```: lists all consumptions done by a ticket
* ```GET "/consumptions/available```: lists all available consumptions
* ```POST "/consumptions/available```: add an available consumption which can be consumed by all tickets
* ```DELETE "/consumptions/available```: removes an available consumption
* ```POST "/consumptions/consume```: consumes a consumption defined by an ```ConsumptionDTO```.
* ```POST "/consumptions/reset```: resets a consumption defined by an ```ConsumptionDTO```.


src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+@RestController
+@PreAuthorize("hasRole('ADMIN')")
+@RequestMapping(value = "/consumptions")
+public class ConsumptionController {
+    @Autowired
+    private ConsumptionService consumptionService;
+
+    @RequestMapping(method = RequestMethod.GET)
+    public Map<String, Boolean> isConsumed(@RequestBody ConsumptionDTO consumptionDTO){
```
skamoen: I think a `@RequestBody` for a `GET` request is a bit weird. I would turn this into `/consumptions/{consumptionId}/tickets/{ticketId}`. More RESTful too

skamoen: Or `/consumptions/{ticketId}` as I'm not sure its required to get a per-consumption overview. Just an overview of all consumptions on a ticket works, just requires some more frond-end logic.

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
```diff
@@ -0,0 +1,81 @@
...
+    public Collection<Consumption> consumptionsMade(@PathVariable Long ticketId){
+        return consumptionService.getByTicketIdIfValid(ticketId).getConsumptionsMade();
+    }
+
+    @RequestMapping(value = "/available", method = RequestMethod.GET)
+    public Collection<Consumption> getAllPossibleConsumptions(){
+        return consumptionService.getPossibleConsumptions();
+    }
+
+    @RequestMapping(value = "/available", method = RequestMethod.POST)
```
skamoen: You can turn this into a `POST` request on `/consumptions` as you're adding one to the big Consumptions collection, the `/available` is not needed.

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
```diff
@@ -0,0 +1,81 @@
...
+    public ResponseEntity<?> addAvailableConsumption(@RequestBody String consumptionName){
+        consumptionService.addPossibleConsumption(consumptionName);
+
+        return createResponseEntity(HttpStatus.OK, "Successfully added " + consumptionName + " as a supported consumption.");
+    }
+
+    @RequestMapping(value = "/available", method = RequestMethod.DELETE)
```
skamoen: Same here. You're not just removing a consumption as available, but from the entire consumption collection

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
```diff
@@ -0,0 +1,81 @@
...
+    public ResponseEntity<?> removeAvailableConsumption(@RequestBody Long consumptionId){
+        Consumption c = consumptionService.removePossibleConsumption(consumptionId);
+
+        return createResponseEntity(HttpStatus.OK, "Successfully removed " + c.getName() + " as a supported consumption.");
+    }
+
+    @RequestMapping(value = "/consume", method = RequestMethod.POST)
```
skamoen: As you're targeting a specific consumption collection belonging to a ticketId, I'd turn this into a `POST` on `/consumptions/{ticketId}/consume`. A post on `/consumptions/consume` is pretty generic for a specific change.

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
```diff
@@ -0,0 +1,81 @@
...
+    public ResponseEntity<?> consume(@RequestBody ConsumptionDTO consumptionDTO){
+        consumptionService.consume(consumptionDTO.getTicketId(), consumptionDTO.getConsumptionId());
+        return createResponseEntity(HttpStatus.OK, "Successfully consumed consumption");
+    }
+
+    @RequestMapping(value = "/reset", method = RequestMethod.POST)
```
skamoen: Same here, you're targetting a specific consumption. A specific URL is more appropriate

src/main/java/ch/wisv/areafiftylan/service/ConsumptionServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+                                  TicketService ticketService) {
+        this.consumptionMapsRepository = consumptionMapsRepository;
+        this.possibleConsumptionsRepository = possibleConsumptionsRepository;
+        this.ticketService = ticketService;
+    }
+
+    @Override
+    public ConsumptionMap getByTicketIdIfValid(Long ticketId) {
+        if(!ticketService.getTicketById(ticketId).isValid()){
+            throw new InvalidTicketException("Ticket is invalid; can't reset consumptions");
```
skamoen: Wrong error message

src/main/java/ch/wisv/areafiftylan/service/ConsumptionServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        Collection<ConsumptionMap> consumptionMaps = consumptionMapsRepository.findAll();
```
skamoen: consumptionMaps is unused

src/main/java/ch/wisv/areafiftylan/service/ConsumptionServiceImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    @Override
+    public boolean isConsumed(Long ticketId, Long consumptionId) {
+        Consumption c = getByConsumptionId(consumptionId);
+        return getByTicketIdIfValid(ticketId).isConsumed(c);
+    }
+
+    private ConsumptionMap InitializeConsumptionMap(Long ticketId){
```
skamoen: Should be lowercase initialize

src/test/java/ch/wisv/areafiftylan/ConsumptionTest.java
```diff
@@ -0,0 +1,424 @@
...
+
+        getTicketConsumptions(ticket).
+            then().
+                statusCode(HttpStatus.SC_OK).
+                body("$", hasSize(1)).
+                body("name", containsInAnyOrder(spicyFood.getName()));
+    }
+
+    @Test
+    public void getTicketConsumptions_OneConsumedAndDeleted(){
```
skamoen: Not sure what this test adds. The service calls don't add anything to get request, as these aren't the calls you're testing.

skamoen:
I'm not a fan of your testing structure. All the private helper methods don't improve the readability, as I'm not sure what they all (exactly do) while reading and they're not independently verified. Furthermore there are a lot of mixed service and repository calls calls, which don't belong in an integration test.

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
bvanderdrift: Could you elaborate why it's better to pass parameters as a PathVariable instead of a DTO? I don't really see the difference, and prefer having all parameters in a single object instead of fragmented over PathVariable/RequestBody.

src/test/java/ch/wisv/areafiftylan/ConsumptionTest.java
bvanderdrift: This one actually threw an error during development which made me find a bug where a ticket would have two ConsumptionMaps. Because of this I'd like to keep it here. It has proven itself!

src/test/java/ch/wisv/areafiftylan/ConsumptionTest.java
```diff
@@ -0,0 +1,424 @@
...
+        consumptionDTO.put("consumptionId", consumptionId.toString());
+        return consumptionDTO;
+    }
+
+    private void invalidateTicket(){
+        ticket.setValid(false);
+        ticketRepository.saveAndFlush(ticket);
+    }
+
+    private Response getAllPossibleTickets_AsAdmin(){
```
bvanderdrift: Wow, totally missed this wrongful naming.

bvanderdrift:
What would belong in a test call? I on the other hand very much prefer the private testing functions. It removes duplicate code, and they have clear method names. It abstracts the calls, but still returns the responses. Maybe a second opinion would be a good idea (although I am not sure who would be suitable).

Could you maybe provide some cases in which you are confused?

bvanderdrift:
Back to @skamoen

src/main/java/ch/wisv/areafiftylan/controller/ConsumptionController.java
skamoen: I guess it's a matter of REST principles vs personal preference. It's more RESTful to target specific objects and have the request body the data that needs to be changed/inserted. I'm fine with this though, as the underlying code doesn't change and it's understandable.

skamoen:
Well, currently it's a mix of unit tests, integration tests and something in between. The idea is that you put something in in the start (The controller) and you check if what comes out is right (the ResponseBody). You can do some additional checks in the repository, but I don't think servicecalls belong in here, as they're part of what you're testing. 

Private methods are fine, but they should purely focus on the request, and you have to validate that they're making the right request and give the right response. I'll look for some concrete examples tomorrow. 

I sort of disagree that duplicate code in these tests is a bad thing, as tests should be fully independent. The RESTassured framework already removes the real duplicate stuff. The way I'd use private methods is re-using previous testcases to come to a certain state. When the state is reached, you want to fully test the complete integration.

bvanderdrift:
So I have chosen to keep the consumptionService in the tests. I completely agree that you perform the tests to test the service as well, but I also think that if a test covers a part of the service, you should be able to use it in further tests. This is because the test proves that part of the service works. If something still goes wrong, that means you haven't tested enough, and not that it's wrong to directly use the service in your tests.

bvanderdrift:
Back to @skamoen

skamoen:
Alright. It's much easier to read what the tests do now and I agree with the service re-use. So LGTM!

PR #242, opened 2016-05-24
CurrentUserController fixes

skamoen:
The CurrentUserController didn't have the `@PreAuthorize` tags on every method, which resulted in some nullpointers. As every method needs it, I moved it to the Controller declaration. 

Also fixed a dataleak on the /current/orders endpoint, where it would show all the information of the ticket owner. This could be problematic for people who transferred their ticket. 

This PR can be squashed

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/controller/CurrentUserRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/controller/CurrentUserRestController.java
@@ -134,7 +134,7 @@ public class CurrentUserRestController {
     }
 
     /**
-     * Gett the seats of the current user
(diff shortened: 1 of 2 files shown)


src/main/java/ch/wisv/areafiftylan/controller/CurrentUserRestController.java
```diff
@@ -150,6 +133,13 @@
         return orderService.getOpenOrders(currentUser.getUsername());
     }
 
+    /**
+     * Gett the seats of the current user
```
bvanderdrift: Get*

bvanderdrift:
LGTM

PR #243, opened 2016-05-26
Token based authentication

skamoen:
Lets try this again (#208), as this is required for Echo-gaming (Intranet)

For clients that are not the browser, the form-based authentication doesn't work. This implements a Token-based authentication method that generates a Token with which clients can perform actions.

Tokens are generated with a POST on /token with a username and password. If credentials are correct, a Token will be returned and saved.

Requests with a token in the X-Auth-Token header will be authenticated by the new Filter.

Requests with a valid token are excluded from CSRF.


skamoen:
I could really use a review on this one, as this will help a lot with testing of the intranet

bvanderdrift:
I'll do it in a minute, is already high on my list for today!

src/main/java/ch/wisv/areafiftylan/util/TestDataRunner.java
```diff
@@ -66,6 +70,12 @@ public void run(String... evt) throws Exception {
         ticketRepository.save(ticket2);
         ticketRepository.save(ticket3);
 
+        Team team = new Team("testTeam", testUser1);
+        team.addMember(testUser2);
+        team.addMember(testUser3);
+
+        teamRepository.save(team);
```
bvanderdrift: Better saveAndFlush?

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
```diff
@@ -61,6 +72,19 @@ public ModelAndView getLoginPage() {
         return createResponseEntity(HttpStatus.OK, "Here's your token!");
     }
 
+    @RequestMapping(value = "/token", method = RequestMethod.POST)
```
bvanderdrift: Just for clarity. This functions as a new login endpoint right?

src/main/java/ch/wisv/areafiftylan/util/TestDataRunner.java
skamoen: It's just the TestDataRunner, it's not essential that's flushed right there. It's been working fine :)

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
skamoen: In a way, yes.

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
```diff
@@ -61,6 +72,19 @@ public ModelAndView getLoginPage() {
...
+    public ResponseEntity<?> createAuthenticationToken(@RequestBody UserDTO userDTO) {
+        String authToken = authenticationService.createNewAuthToken(userDTO.getUsername(), userDTO.getPassword());
+
+        return createResponseEntity(HttpStatus.OK, "Token successfully created", authToken);
+    }
+
+    @PreAuthorize("isAuthenticated()")
+    @RequestMapping(value = "/token/verify", method = RequestMethod.POST)
+    public ResponseEntity<?> verifyToken() {
```
bvanderdrift: This method has been created under the assumption that if someone reaches this endpoint the token is valid by default? Maybe it's better to send the token as well, and actually perform a verification on the actual token to be fully sure?

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
```diff
@@ -61,6 +72,19 @@ public ModelAndView getLoginPage() {
...
+        return createResponseEntity(HttpStatus.OK, "Token is valid!");
```
bvanderdrift: If this endpoint is a POST, it seems an action happens. That a token actually gets verified. From reading the response it seems this is just getting the validity of a token. No action happens upon reaching this endpoint. I think this should be a GET endpoint.

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
skamoen: The token is checked in the Filter. If it's not valid, there won't be an Authenticated user so the `PreAuthorize` will catch it.

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
skamoen: Good point. I agree.

bvanderdrift:
Just for understanding. If I understand this correctly this does not replace CSRF, but adds Token authentication on top of CSRF, with CSRF as fallback if no token is defined?

src/main/java/ch/wisv/areafiftylan/security/TokenAuthenticationFilter.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+
+        String xAuth = ((HttpServletRequest) request).getHeader("X-Auth-Token");
+
+        if (!Strings.isNullOrEmpty(xAuth)) {
+            AuthenticationToken authenticationToken =
+                    extractOptional(authenticationTokenRepository.findByToken(xAuth), response);
+            if (authenticationToken.isValid()) {
```
bvanderdrift: Here a NullPointer can be thrown. The extractOptional method can return null.

skamoen:
It does replace CSRF. CSRF is still here because the form-based login still exists. The CSRF filter disables CSRF when a Token is present in the header. If there's no Token in the header, the regular authentication methods apply (including csrf)

src/main/java/ch/wisv/areafiftylan/security/TokenAuthenticationFilter.java
skamoen: The sendError method interrupts the chain, so it won't continue with a null value, similar to how an exception works.

src/main/java/ch/wisv/areafiftylan/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+            authenticationTokenRepository.findByUserUsername(username)
+                    .ifPresent(t -> authenticationTokenRepository.delete(t));
+
+            return authenticationTokenRepository.save(new AuthenticationToken(user, 60 * 24 * 2)).getToken();
+        }
+
+        throw new AuthenticationCredentialsNotFoundException("Incorrect credentials");
+    }
+
+    private boolean correctCredentials(User user, String password) {
```
bvanderdrift: I feel like this already has been implemented in the Spring framework. Is it not a better idea to use their checks?

src/main/java/ch/wisv/areafiftylan/security/token/AuthenticationToken.java
```diff
@@ -0,0 +1,20 @@
...
+
+import ch.wisv.areafiftylan.model.User;
+
+import javax.persistence.Entity;
+
+/**
+ * Created by Sille Kamoen on 6-5-16.
+ */
+@Entity
+public class AuthenticationToken extends Token {
```
bvanderdrift: Please use a final EXPIRATION field for default expiration length.

Can be seen in https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/security/token/Token.java#L14

src/main/java/ch/wisv/areafiftylan/service/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public String createNewAuthToken(String username, String password) {
+        User user = userService.getUserByUsername(username).orElseThrow(() -> new UsernameNotFoundException(username));
+
+        if (correctCredentials(user, password)) {
+            // Delete the old Token
+            authenticationTokenRepository.findByUserUsername(username)
+                    .ifPresent(t -> authenticationTokenRepository.delete(t));
+
+            return authenticationTokenRepository.save(new AuthenticationToken(user, 60 * 24 * 2)).getToken();
```
bvanderdrift: I see a magic number! Please use an internal EXPIRATION final as seen in [other tokens](https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/security/token/Token.java#L14) for default values.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+            content(userDTO).contentType(ContentType.JSON).
+            post("/token").
+        then().
+            statusCode(HttpStatus.SC_OK);
+        //@formatter:on
+
+        AuthenticationToken authenticationToken2 =
+                authenticationTokenRepository.findByUserUsername(user.getUsername()).orElse(null);
+
+        Assert.assertFalse(authenticationToken1.equals(authenticationToken2));
```
bvanderdrift: I don't see a 'equals' implementation in both AuthenticationToken and the Token class itself. If I'm correct this means that it checks for object equality (similar to ==) which will always throw false if comparing two repo results.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Test
+    public void testRequestTokenRenew() {
```
bvanderdrift: I feel like this test and the above test have switched titles.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+        userDTO.put("username", user.getUsername());
+        userDTO.put("password", userCleartextPassword);
+
+        //@formatter:off
+        given().
+        when().
+            content(userDTO).contentType(ContentType.JSON).
+            post("/token").
+        then().
+            statusCode(HttpStatus.SC_OK);
```
bvanderdrift: I personally would also extract the token in the response and check it with the token from the repository. Seems like a very weird fail if you look into the code, but can't harm.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+        when().
+            post("/token/verify").
+        then().
+            statusCode(HttpStatus.SC_OK);
+        //@formatter:on
+    }
+
+    @Test
+    public void testVerifyInvalidToken() {
+        Header header = new Header("X-Auth-Token", "invalid");
```
bvanderdrift: I personally would have stored this header key in a private final in this class. These are magic strings. Same goes for the invalid header.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+        //@formatter:off
+        given().
+            header(header).
+        when().
+            post("/token/verify").
+        then().
+            statusCode(HttpStatus.SC_FORBIDDEN);
+        //@formatter:on
+    }
+
```
bvanderdrift: Also add a test with a missing header?

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
```diff
@@ -0,0 +1,225 @@
...
+    public void testVerifyMissingToken() {
+        Header header = new Header("X-Auth-Token", "");
+
+        //@formatter:off
+        given().
+            header(header).
+        when().
+            post("/token/verify").
+        then().
+            statusCode(HttpStatus.SC_FORBIDDEN);
```
bvanderdrift: I would a expect a bad request here.

src/main/java/ch/wisv/areafiftylan/controller/AuthenticationController.java
```diff
@@ -34,17 +37,25 @@
 @Log4j2
 public class AuthenticationController {
 
-    @Autowired
-    MailService mailService;
+    private MailService mailService;
```
bvanderdrift: Mailservice is unused

bvanderdrift:
Looking good! Minor comments but overall a good addition in my opinion. I am not 100% confident in my knowledge on Spring security, but this has to do for now.

bvanderdrift:
Back to @skamoen

src/main/java/ch/wisv/areafiftylan/service/AuthenticationServiceImpl.java
skamoen: I wouldn't know where or how. It's integrated into the entire authentication system, but that's not what we want here. It's quite straightforward right?

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
skamoen: With a missing header, it's the same as a generic anonymous request, which is tested in all the other tests.

src/test/java/ch/wisv/areafiftylan/TokenAuthenticationTest.java
skamoen: I consider this equal to a request without header, which would also result in a FORBIDDEN. I don' think an header error should result in a bad request status.

skamoen:
I answered or fixed all your points, back to @bvanderdrift

bvanderdrift:
LGTM!

PR #247, opened 2016-05-29
Update sample properties

skamoen:
This PR provides a small update to the sample application properties to explain some unclear differences between production and dev profiles.


TimvdLippe:
LGTM

PR #257, opened 2016-06-03
Remove auth on seatmap load

svenpopping:


martijn9612:
The anon test is removed, the content is already visible upon creating an account.

PR #284, opened 2016-10-02
Upgrade Spring Boot to 1.4.1

skamoen:
Upgrade all the things!

Had to change some fields to be compatible with the new Hibernate 5.0. There are many new features we could use, especially in the testing department.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/users/model/User.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/model/User.java
@@ -25,8 +25,8 @@ import org.springframework.security.core.userdetails.UserDetails;
 
 import javax.persistence.*;
 import java.io.Serializable;
-import java.util.Collection;
 import java.util.HashSet;
+import java.util.Set;
 
 
 @Entity
@@ -57,7 +57,7 @@ public class User implements Serializable, UserDetails {
     @ElementCollection(targetClass = Role.class, fetch = FetchType.EAGER)
     @Enumerated(EnumType.STRING)
     @CollectionTable(name = "user_role")
-    private Collection<Role> roles;
+    private Set<Role> roles;
 
     @JsonIgnore
     private boolean accountNonExpired = true;
@@ -93,7 +93,7 @@ public class User implements Serializable, UserDetails {
     }
 
     @Override
-    public Collection<? extends GrantedAuthority> getAuthorities() {
+    public Set<? extends GrantedAuthority> getAuthorities() {
         return roles;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/ch/wisv/areafiftylan/extras/consumption/model/ConsumptionMap.java
```diff
@@ -37,7 +38,7 @@
 
     @NonNull
     @ManyToMany(fetch = FetchType.EAGER)
-    private Collection<Consumption> consumptionsMade;
+    private List<Consumption> consumptionsMade;
```
bvanderdrift: Could you shed some light on the change?

src/main/java/ch/wisv/areafiftylan/users/model/User.java
```diff
@@ -57,7 +58,7 @@
     @ElementCollection(targetClass = Role.class, fetch = FetchType.EAGER)
     @Enumerated(EnumType.STRING)
     @CollectionTable(name = "user_role")
-    private Collection<Role> roles;
+    private Set<Role> roles;
```
bvanderdrift: Could you shed some light on this alteration?

src/main/java/ch/wisv/areafiftylan/extras/consumption/model/ConsumptionMap.java
mgajanssen: In my opinion this is ok, an more abstract class allows for more kinds of internal return types, allowing for more optimization. This will not change the data returned, it will still be an array-like return.

skamoen: Hibernate 5.0 doesn't like initializing 2 collections at once. By changing this to more specific implementations of a Collection, it was able to do so. Not exactly sure what the internal difference is, but  considering the nature of the collections, this doesn't hurt

mgajanssen: Whoops, my bad, this is not the case. I, too, would be interested.

bvanderdrift: @skamoen So now here a list is instantiated and somewhere else a collection? What if you change that collection to a list? Doesn't the ambiguity return?

skamoen: Fair point. I reversed the consumptionsMade to Collection, as changing just the Roles to a Set seems to do the trick. I checked for all occurences, no problems found.

PR #287, opened 2016-10-24
Implement new mailsending methods

### LINKED ISSUE #281: Mail rework
Endpoints to perform specific actions
- [ ] specific users
- [ ] teams
- [ ] users with ticket
- [ ] users without ticket
- [ ] users with pickup
- [ ] seatgroups
- [ ] manual

mgajanssen:
Fixes #281

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/users/service/UserRepository.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/service/UserRepository.java
@@ -28,6 +28,4 @@ import java.util.Optional;
 public interface UserRepository extends JpaRepository<User, Long> {
 
     Optional<User> findOneByUsernameIgnoreCase(String username);
-
-    Collection<User> findByPickupTrue();
 }
--- a/src/main/java/ch/wisv/areafiftylan/users/service/UserRepository.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/service/UserRepository.java
@@ -21,10 +21,13 @@ import ch.wisv.areafiftylan.users.model.User;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.util.Collection;
 import java.util.Optional;
 
 @Repository
 public interface UserRepository extends JpaRepository<User, Long> {
 
     Optional<User> findOneByUsernameIgnoreCase(String username);
+
+    Collection<User> findByPickupTrue();
 }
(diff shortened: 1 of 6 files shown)


src/main/java/ch/wisv/areafiftylan/utils/mail/MailRestController.java
```diff
@@ -64,9 +76,60 @@ public MailRestController(MailService mailService, UserService userService, Team
...
+                .map(Ticket::getOwner)
+                .collect(Collectors.toList());
+
+        mailService.sendTemplateMailToAll(users, mailDTO);
+
+        return createResponseEntity(HttpStatus.OK, "Mail was successfully sent");
+    }
+
+    @RequestMapping(value = "/user/{seatgroup}", method = RequestMethod.POST)
+    ResponseEntity<?> sendMailToSeatGroup(@RequestParam String seatGroup, @Validated @RequestBody MailDTO mailDTO) {
```
skamoen: Why is this useful? Can't really think of a realistic case where we want to mail specific seatgroups

src/main/java/ch/wisv/areafiftylan/users/service/UserRepository.java
```diff
@@ -21,6 +21,7 @@
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
+import java.util.Collection;
```
mgajanssen: I think this collection does not have to be in here. It is never used.

src/main/java/ch/wisv/areafiftylan/utils/mail/MailRestController.java
```diff
@@ -28,24 +32,32 @@
...
 @PreAuthorize("hasRole('ADMIN')")
 @RequestMapping("/mail")
 public class MailRestController {
 
+
+    private MailService mailService;
+    private UserService userService;
+    private TeamService teamService;
+    private TicketService ticketService;
+    private SeatService seatService;
```
mgajanssen: The seatservice can also be removed, the field is not used.

mgajanssen:
It would also be very useful to be able to email all the users who have not yet reserved a seat.

skamoen:
We can do two things. Either merge this and schedule a replacement, or not merge this and keep it as backup if we need this before a replacement is done... Both are fine with me, up to @mgajanssen

mgajanssen:
I think it is good to keep this as a backup and not merge it. Lets find a better solution first.

skamoen:
Closing, but keeping the branch.

PR #288, opened 2016-10-29
Add birthday field

### LINKED ISSUE #151: Add birthday field in profile
This is needed for the alcohol permitation.

mgajanssen:
Fixes #151 

We can now check the age of the participants using the age stored in the database.


skamoen:
That was easier then expected. LGTM!

PR #292, opened 2016-11-06
Add contact form

### LINKED ISSUE #291: Contact form
As of now, there is an email address on the website, this accumulates a lot of spam on that email-address. A contact form would be a good  way to solve this problem.

martijnjanssen:
Fixes #291 
This allows us to remove all the spam we reveive and make it easier for our users to approach us.


src/main/java/ch/wisv/areafiftylan/utils/mail/MailDTO.java
```diff
@@ -19,14 +19,16 @@
 
 import lombok.Getter;
 import lombok.Setter;
+import org.hibernate.validator.constraints.NotEmpty;
 
-public class MailDTO {
+import javax.validation.constraints.NotNull;
```
TimvdLippe: This one is unused?

PR #294, opened 2016-11-08
Add caching in between builds

martijnjanssen:
According to the travis manual [this](https://docs.travis-ci.com/user/languages/java/#Caching) is the correct way to cache gradle dependencies.


skamoen:
image

PR #296, opened 2016-11-13
Remove false duplicate displayname when not changed in update function

martijnjanssen:


src/main/java/ch/wisv/areafiftylan/users/service/UserServiceImpl.java
```diff
@@ -149,7 +149,10 @@ public User edit(Long userId, UserDTO userDTO) {
     public User addProfile(Long userId, ProfileDTO profileDTO) throws DataIntegrityViolationException {
         // No two identical displayNames are allowed
         userRepository.findOneByProfileDisplayNameIgnoreCase(profileDTO.getDisplayName()).ifPresent(u -> {
-            throw new DataIntegrityViolationException("DisplayName already in use");});
+            if (u.getId() != userId) {
+                throw new DataIntegrityViolationException("DisplayName already in use");
```
skamoen: Why are you doing it here in a questionable way, instead of a database constraint? This doesn't look like proper usage of the `ifPresent()`

PR #297, opened 2016-11-13
Add timezone to the date & fix zero based month

martijnjanssen:


src/main/resources/config/application.properties.sample
```diff
@@ -72,13 +71,13 @@ logging.config=classpath:log4j2-prod.xml
 # Key for sending a paymentrequest to the Mollie api
 a5l.molliekey=[CHANGE ME]
 # The return url for the frontend that displays the success message
-a5l.paymentReturnUrl=https://areafiftylan.nl/#!/ordersuccess
+a5l.paymentReturnUrl=https://areafiftylan.nl/ordersuccess
```
TimvdLippe: 😂

src/main/resources/config/application-test.properties
```diff
@@ -42,15 +42,16 @@ spring.thymeleaf.content-type=text/html
 
 # The date format we use
 spring.jackson.date-format=yyyy-MM-dd
+spring.jackson.time-zone=GMT+1
```
skamoen: Did you check if this is the correct timezone, which also deals correctly with daylight savings?

PR #298, opened 2016-11-17
Upgrade Mollie API-client to 2.3.0

### LINKED ISSUE #290: Upgrade Mollie API
A new major version of the Mollie API was released, so we should upgrade. There are some breaking changes. See https://github.com/WISVCH/payments/blob/master/src/main/java/ch/wisv/payments/rest/MolliePaymentService.java for a working example.

skamoen:
We were still using 1.0.1. It involved some syntax changes, but there are some good changes which made it worth upgrading. 

Fixes #290


martijnjanssen:
Looks good! Thanks!

PR #299, opened 2016-11-21
Add Icon ignore

FrancisBehnen:
Don't ask why this is needed.... it just is :/


FrancisBehnen:
Added to global gitignore instead

PR #301, opened 2016-11-21
External html templates

### LINKED ISSUE #258: Add external mailtemplate support
Html templates can be loaded from an external location. If there's a file present in `/config/templates/`, this should be used as emailtemplate instead.

skamoen:
Moved templates outside of classpath so they can be configurable.

Fixes #258


martijnjanssen:
Are the `application.properties` still supposed to be in the resources folder?

skamoen:
I moved the application.properties outside of the build as well. This way, when running locally the build locally you can use the properties in the location where they are now, but a built jar file won't contain your applicaiton.properties, as they should be located in a `./config` folder. It's safer this way.

PR #304, opened 2016-11-21
Add requesthandler to log out & delete token

martijnjanssen:
This implements a new logout endpoint to de-activate a token. It takes the current token from the header and de-activates it immediately.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/exception/TokenNotFoundException.java
+++ b/src/main/java/ch/wisv/areafiftylan/exception/TokenNotFoundException.java
@@ -19,7 +19,7 @@ package ch.wisv.areafiftylan.exception;
 
 public class TokenNotFoundException extends RuntimeException {
 
-    public TokenNotFoundException(String token) {
-        super("could not find token '" + token + "'.");
+    public TokenNotFoundException(String message) {
+        super(message);
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/exception/TokenNotFoundException.java
+++ b/src/main/java/ch/wisv/areafiftylan/exception/TokenNotFoundException.java
@@ -19,7 +19,7 @@ package ch.wisv.areafiftylan.exception;
 
 public class TokenNotFoundException extends RuntimeException {
 
-    public TokenNotFoundException(String message) {
-        super(message);
+    public TokenNotFoundException(String token) {
+        super("Could not find token " + token);
     }
 }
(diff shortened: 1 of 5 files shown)


FrancisBehnen:
This is related with: https://github.com/AreaFiftyLAN/lancie-frontend/issues/275 right? Is there a corresponding issue in this repo?

martijnjanssen:
This is not related to that issue, this is just added functionality to allow our users to log out and expire the token at an earlier point. As of now the token just expired after a specified duration, now the users have the possibility to actively log out.

src/main/java/ch/wisv/areafiftylan/exception/TokenNotFoundException.java
```diff
@@ -19,7 +19,7 @@
 
 public class TokenNotFoundException extends RuntimeException {
 
-    public TokenNotFoundException(String token) {
-        super("could not find token '" + token + "'.");
+    public TokenNotFoundException(String message) {
+        super(message);
```
skamoen: This change "breaks" all previous usage of this exception.

src/main/java/ch/wisv/areafiftylan/exception/XAuthTokenNotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package ch.wisv.areafiftylan.exception;
+
+public class XAuthTokenNotFoundException extends TokenNotFoundException {
+    public XAuthTokenNotFoundException () {
+        super("X-Auth-Token not found");
```
skamoen: If this is the only use case, it can directly extend `RuntimeException`.

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationController.java
```diff
@@ -97,6 +105,25 @@ public ModelAndView getLoginPage() {
         return createResponseEntity(HttpStatus.OK, "Token is valid!");
     }
 
+    @PreAuthorize("isAuthenticated()")
+    @RequestMapping(value = "/token/logout", method = RequestMethod.GET)
```
skamoen: Sounds like a weird path for (the only) logout functionality

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationController.java
```diff
@@ -97,6 +105,25 @@ public ModelAndView getLoginPage() {
...
+    public ResponseEntity<?> removeSession(@RequestHeader("X-Auth-Token") String xAuth) {
+        if (!Strings.isNullOrEmpty(xAuth)) {
+            Optional<AuthenticationToken> tokenOptional = tokenRepository.findByToken(xAuth);
+            AuthenticationToken token = tokenOptional.orElseThrow(XAuthTokenNotFoundException::new);
+            if (!token.isValid()) {
```
skamoen: With this amount of logic, it may finally be time for a dedicated `AuthenticationService`

PR #306, opened 2016-11-24
IntelliJ Code Analysis cleanup

skamoen:
Based in IntelliJ's code inspection, this PR introduces hundreds of small fixes that improve the SEM-factor.


TimvdLippe:
In general seems fine, but when I run `Analyze -> Inspect Code` I get a lot more warnings. Could you also fix these?

skamoen:
I didn't improve the test sources, as these will be overhauled soon enough anyway

PR #308, opened 2016-11-26
Add dev flag for bootRun to populate database

### LINKED ISSUE #307: Add documentation on how to run server from the command line
Right now I do not need to edit the server, I just need to run it. However, intellij is a significant overhead and I would like to prevent that. Please include instructions in the README on how to run the server locally.

Note that `./gradlew bootRun` does run the server, but not with the dev profile (and therefore has no users etc...)

martijnjanssen:
Fixes #307


martijnjanssen:
I have now made a seperate task for the bootRun with the settings.

It is also possible to add a flag, but this is an easier solution for the user and does exactly the same.

Issue #262, opened 2016-07-01
Dynamic ticket options


Icety:
You need to be able to generate new ticket options in the backend.
Every option should consist of a name/title and a monetarian value.
ie. 
- Pickupservice (€2,50)
- CH-Member (- €5,-)

PR #312, opened 2016-12-02
Feature: Anonymous orders

### LINKED ISSUE #271: Anonymous orders
An order should be created without a user when there is no user logged in.

### LINKED ISSUE #272: You should be able to couple a user to an anonymous order
When you checkout with an anonymous order, there should be functionality to couple a user to the order.

### LINKED ISSUE #282: Order id added to register
When registering while checking out, it should be possible to add the orderId to the register action. When sending the confirmation email, the orderId should be added as a parameter.

### LINKED ISSUE #302: Remove Role from UserDTO
potential security issues

skamoen:
This PR introduces the new Order flow for A5L 2017. 

Orders are created anonymous by default, at which point tickets can be added by anyone. Before checkout, the Order should be assigned to a User, after which the old functionality applies. 

A new methodology is introduced for the testing of Services. It follows a more Unit-like testing method, resulting in a huge speed increase and better coverage. 

The OrderRestIntegrationTest is sub-optimal, but as #262 is going to be implemented soon, I didn't bother with a detailed overhaul. 

Fixes #271 
Fixes #272 
Fixes #282 
Fixes #302

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
@@ -23,7 +23,6 @@ import ch.wisv.areafiftylan.exception.TicketUnavailableException;
 import ch.wisv.areafiftylan.products.model.Order;
 import ch.wisv.areafiftylan.products.model.TicketDTO;
 import ch.wisv.areafiftylan.products.service.OrderService;
-import ch.wisv.areafiftylan.users.model.User;
 import ch.wisv.areafiftylan.utils.view.View;
 import com.fasterxml.jackson.annotation.JsonView;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -65,17 +64,14 @@ public class OrderRestController {
      * always contains at least one ticket. Optional next tickets should be added to the order by POSTing to the
      * location provided.
      *
-     * @param auth      The User that is currently logged in
      * @param ticketDTO Object containing information about the Ticket that is being ordered.
      *
      * @return A message informing about the result of the request
      */
-    @PreAuthorize("isAuthenticated()")
     @RequestMapping(value = "/orders", method = RequestMethod.POST)
     @JsonView(View.OrderOverview.class)
-    public ResponseEntity<?> createOrder(Authentication auth, @RequestBody @Validated TicketDTO ticketDTO) {
+    public ResponseEntity<?> createOrder(@RequestBody @Validated TicketDTO ticketDTO) {
         HttpHeaders headers = new HttpHeaders();
-        User user = (User) auth.getPrincipal();
 
         // You can't buy non-buyable Tickts for yourself, this should be done via the createAdminOrder() method.
         if (!ticketDTO.getType().isBuyable()) {
@@ -83,24 +79,10 @@ public class OrderRestController {
                     "Can't order tickets with type " + ticketDTO.getType().getText());
         }
 
-        Order order = orderService.create(user.getId(), ticketDTO);
+        Order order = orderService.create(ticketDTO);
 
         headers.setLocation(
                 ServletUriComponentsBuilder.fromCurrentRequest().path("/{id}").buildAndExpand(order.getId()).toUri());
-
-        return createResponseEntity(HttpStatus.CREATED, headers,
-                "Ticket available and order successfully created at " + headers.getLocation(), order);
-    }
-
-    @PreAuthorize("hasRole('ADMIN')")
-    @JsonView(View.OrderOverview.class)
-    @RequestMapping(value = "/users/{userId}/orders", method = RequestMethod.POST)
-    public ResponseEntity<?> createAdminOrder(@PathVariable Long userId, @RequestBody @Validated TicketDTO ticketDTO) {
-        HttpHeaders headers = new HttpHeaders();
-        Order order = orderService.create(userId, ticketDTO);
-
-        headers.setLocation(ServletUriComponentsBuilder.fromCurrentContextPath().path("/orders/{id}").
-                buildAndExpand(order.getId()).toUri());
 
         return createResponseEntity(HttpStatus.CREATED, headers,
                 "Ticket available and order successfully created at " + headers.getLocation(), order);
@@ -154,6 +136,24 @@ public class OrderRestController {
     }
 
     /**
+     * Assign a User to a previously anonymous Order. The currently assigned user is assigned to the order specified in
+     * the path. This is required before an order can be paid.
+     *
+     * @param auth    The Spring Authentication object
+     * @param orderId Id of the Order to be assigned to the currently logged in user
+     *
+     * @return The assigned order
+     */
+    @PreAuthorize("isAuthenticated() and @currentUserServiceImpl.canAccessOrder(principal, #orderId)")
+    @PostMapping(value = "/orders/{orderId}/assign")
+    @JsonView(View.OrderOverview.class)
+    public ResponseEntity<?> assignOrderToUser(Authentication auth, @PathVariable Long orderId) {
+        Order attachedOrder = orderService.assignOrderToUser(orderId, auth.getName());
+        return createResponseEntity(HttpStatus.OK, "Order successfully attached to User", attachedOrder);
+    }
+
+
+    /**
      * This method requests payment of the order, locks the order and needs to return information on how to proceed.
      * Depending on PaymentService.
      *
@@ -198,7 +198,6 @@ public class OrderRestController {
      */
     @RequestMapping(value = "/orders/status", method = RequestMethod.POST)
     public ResponseEntity<?> updateOrderStatus(@RequestParam(name = "id") String orderReference) {
-        //TODO: Figure out how Mollie sends this request
         orderService.updateOrderStatus(orderReference);
         return createResponseEntity(HttpStatus.OK, "Status is being updated");
     }
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
@@ -132,7 +132,9 @@ public class OrderRestController {
     @RequestMapping(value = "/orders/{orderId}", method = RequestMethod.DELETE)
     @JsonView(View.OrderOverview.class)
     public ResponseEntity<?> removeFromOrder(@PathVariable Long orderId, @RequestBody @Validated TicketDTO ticketDTO) {
-        Order modifiedOrder = orderService.removeTicketFromOrder(orderId, ticketDTO);
+        Order modifiedOrder = orderService
+                .removeTicketFromOrder(orderId, ticketDTO.getType(), ticketDTO.hasPickupService(),
+                        ticketDTO.isCHMember());
         return createResponseEntity(HttpStatus.OK, "Ticket successfully removed from Order", modifiedOrder);
     }
 
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
@@ -79,7 +79,7 @@ public class OrderRestController {
                     "Can't order tickets with type " + ticketDTO.getType().getText());
         }
 
-        Order order = orderService.create(ticketDTO);
+        Order order = orderService.create(ticketDTO.getType(), ticketDTO.hasPickupService(), ticketDTO.isCHMember());
 
         headers.setLocation(
                 ServletUriComponentsBuilder.fromCurrentRequest().path("/{id}").buildAndExpand(order.getId()).toUri());
@@ -115,7 +115,8 @@ public class OrderRestController {
     @RequestMapping(value = "/orders/{orderId}", method = RequestMethod.POST)
     @JsonView(View.OrderOverview.class)
     public ResponseEntity<?> addToOrder(@PathVariable Long orderId, @RequestBody @Validated TicketDTO ticketDTO) {
-        Order modifiedOrder = orderService.addTicketToOrder(orderId, ticketDTO);
+        Order modifiedOrder = orderService
+                .addTicketToOrder(orderId, ticketDTO.getType(), ticketDTO.hasPickupService(), ticketDTO.isCHMember());
         return createResponseEntity(HttpStatus.OK, "Ticket successfully added to your order", modifiedOrder);
     }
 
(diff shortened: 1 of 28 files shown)


src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
```diff
@@ -149,11 +132,31 @@ public Order getOrderById(@PathVariable Long orderId) {
...
+     * @param auth    The Spring Authentication object
+     * @param orderId Id of the Order to be assigned to the currently logged in user
+     *
+     * @return The assigned order
+     */
+    @PreAuthorize("isAuthenticated() and @currentUserServiceImpl.canAccessOrder(principal, #orderId)")
+    @PostMapping(value = "/orders/{orderId}/assign")
+    @JsonView(View.OrderOverview.class)
+    public ResponseEntity<?> assignOrderToUser(Authentication auth, @PathVariable Long orderId) {
+        Order assignedOrder = orderService.assignOrderToUser(orderId, auth.getName());
```
martijnjanssen: Any reason why to pick the name of the user instead of the `Id`?

src/main/java/ch/wisv/areafiftylan/products/model/Ticket.java
```diff
@@ -29,75 +31,59 @@
...
-    }
-
-    public void setOwner(User owner) {
-        this.owner = owner;
-    }
-
-    public TicketType getType() {
-        return type;
-    }
-
```
martijnjanssen: Nice job cleaning up!

src/main/java/ch/wisv/areafiftylan/products/service/OrderService.java
```diff
@@ -35,35 +35,42 @@
...
      *
-     * @param orderId   Order from which the Ticket should be removed
-     * @param ticketDTO Ticket which should be removed from the Order
+     * @param orderId       The Id of the Order from which the tickets have to be removed
+     * @param type          The Type of the ticket to be removed
+     * @param pickupService The pickupservice bool of the ticket to be removed
+     * @param chMember      The chMember bool of the ticket to be removed
      *
-     * @return The modified Order
+     * @return The updated Order
```
martijnjanssen: `@return The updated Order where ticket has been removed if present`?

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -76,38 +75,34 @@ public Order getOrderById(Long id) {
...
 
     @Override
-    public Order addTicketToOrder(Long orderId, TicketDTO ticketDTO) {
-        Order order = orderRepository.getOne(orderId);
+    public Order addTicketToOrder(Long orderId, TicketType type, boolean pickupService, boolean chMember) {
+        Order order = getOrderById(orderId);
 
         // Check Order status
-        if (!order.getStatus().equals(OrderStatus.CREATING)) {
+        if (!order.getStatus().equals(OrderStatus.ANONYMOUS) && !order.getStatus().equals(OrderStatus.ASSIGNED)) {
```
martijnjanssen: I would change the condition to 
```java
!(order.getStatus().equals(OrderStatus.ANONYMOUS) || order.getStatus().equals(OrderStatus.ASSIGNED))
```

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -116,25 +111,41 @@ public Order addTicketToOrder(Long orderId, TicketDTO ticketDTO) {
...
+    public Order removeTicketFromOrder(Long orderId, TicketType type, boolean pickupService, boolean chMember) {
+        Order order = getOrderById(orderId);
+        if (order.getStatus().equals(OrderStatus.ANONYMOUS) || order.getStatus().equals(OrderStatus.ASSIGNED)) {
 
             // Find a Ticket in the order, equal to the given DTO. Throw an exception when the ticket doesn't exist
-            Ticket ticket = order.getTickets().stream().filter(isEqualToDTO(ticketDTO)).findFirst()
-                    .orElseThrow(TicketNotFoundException::new);
+            Ticket ticket =
+                    order.getTickets().stream().filter(isEqualToInput(type, pickupService, chMember)).findFirst()
+                            .orElseThrow(TicketNotFoundException::new);
```
martijnjanssen: I would place every action you do after `.stream()` on a new line, increasing the readability.
```java
Ticket ticket = order.getTickets().stream()
    .filter(isEqualToInput(type, pickupService, chMember))
    .findFirst()
    .orElseThrow(TicketNotFoundException::new)
```

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -116,25 +111,41 @@ public Order addTicketToOrder(Long orderId, TicketDTO ticketDTO) {
...
 
             order.getTickets().remove(ticket);
             ticketService.removeTicket(ticket.getId());
```
martijnjanssen: Do we want to make the orderService take care of this? This is an action required when removing a ticket from every order. A separate call for this looks quite wierd. I know this is already in here, but it still caught my attention.

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -168,16 +185,21 @@ public Order updateOrderStatus(String orderReference) {
     }
 
     private void validateTicketsIfPaid(Order order) {
+        if (order.getUser() == null) {
+            throw new UnassignedOrderException(order.getId());
+        }
+
         if (order.getStatus().equals(OrderStatus.PAID)) {
             for (Ticket ticket : order.getTickets()) {
+                ticketService.assignTicketToUser(ticket.getId(), order.getUser().getUsername());
```
martijnjanssen: I would perform this action when assigning the order to the user. Since that is the point from where on the order, and tickets are assigned to a user. From that point on the user can start adding tickets under his own account.

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
```diff
@@ -198,6 +202,15 @@ public void cancelTicketTransfer(String token) {
         return ticketRepository.findByPickupService_True();
     }
 
+    @Override
+    public Ticket assignTicketToUser(Long ticketId, String username) {
+        Ticket ticket = getTicketById(ticketId);
+        User user = userService.getUserByUsername(username);
+        ticket.setOwner(user);
+        return ticketRepository.save(ticket);
```
martijnjanssen: This ticket is not immediately flushed to the database, when does this happen? I can't remember having specific flush calls at the end of a method.

src/main/java/ch/wisv/areafiftylan/users/service/UserServiceImpl.java
```diff
@@ -70,8 +69,9 @@ public User getUserById(long id) {
     }
 
     @Override
-    public Optional<User> getUserByUsername(String username) {
-        return userRepository.findOneByUsernameIgnoreCase(username);
+    public User getUserByUsername(String username) {
+        return userRepository.findOneByUsernameIgnoreCase(username)
+                .orElseThrow(() -> new UsernameNotFoundException("User '" + username + "' not found"));
```
martijnjanssen: Very nice alteration, I was wondering how you were able to remove the throw statement in TeamServiceImpl.

src/test/java/ch/wisv/areafiftylan/IntegrationTest.java
```diff
@@ -127,6 +125,21 @@ public void tearDownIntegrationTest() {
         RestAssured.reset();
     }
 
+
+    protected Header getCSRFHeader() {
```
martijnjanssen: Don't you want to change the authentication in the tests already here? CSRF is going to be disabled at some point in time.

src/test/java/ch/wisv/areafiftylan/products/service/OrderServiceTest.java
```diff
@@ -0,0 +1,726 @@
...
+        thrown.expect(OrderNotFoundException.class);
+        thrown.expectMessage("Order with id: " + id + " not found");
+
+        orderService.getOrderById(id);
+    }
+
+    @Test
+    public void getOrderByIdNull() {
+        thrown.expect(OrderNotFoundException.class);
+        thrown.expectMessage("Order with id: ");
```
martijnjanssen: Shouldn't you add the `" not found"` string at the end?

martijnjanssen:
One other thing, currently with this PR, the scheduler that takes care of cleaning up orders throws an exception because there is no user set on the order.

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
skamoen: Because I'm not sure if the Spring `auth` object has the proper userId when casted.

src/test/java/ch/wisv/areafiftylan/products/service/OrderServiceTest.java
skamoen: the expectMessage function checks for a substring.

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
```diff
@@ -149,11 +132,31 @@ public Order getOrderById(@PathVariable Long orderId) {
...
+        return createResponseEntity(HttpStatus.OK, "Order successfully attached to User", assignedOrder);
```
martijnjanssen: Wouldn't it be better to return the updated order here? Changes are made to the order changing behavior with a POST request. It would seem logical to me that I get the updated order as a response.

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
skamoen: This has to do with the one-way relation between Orders and Tickets. This can't be done automatically unless a Ticket gets a referenced order. At least that was my conclusion last time I looked at it.

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
skamoen: As this method is called at least once per paid order, you're sure this happens once for each ticket. When a ticket is invalid, it doesn't really need a user anyway.

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
skamoen: As previously discussed, it does.

src/main/java/ch/wisv/areafiftylan/users/model/UserDTO.java
```diff
@@ -35,8 +33,7 @@
     @Setter
     private String password = "";
 
-    @NotNull
     @Getter
     @Setter
```
martijnjanssen: I assume this is not needed, it is never used.

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
martijnjanssen: Would this be something to test or do you prefer to do it like this?

src/main/java/ch/wisv/areafiftylan/products/model/Order.java
```diff
@@ -34,89 +36,72 @@
...
-        //JPA only
-    }
-
-    public Long getId() {
-        return id;
-    }
-
-    public Set<Ticket> getTickets() {
-        return tickets;
+    public Order(User user) {
```
martijnjanssen: Where are we still using this for? I thought we wanted to have one path in creating orders and not two separate ones?

src/main/java/ch/wisv/areafiftylan/products/model/Order.java
```diff
@@ -34,89 +36,72 @@
...
         float price = 0F;
         for (Ticket ticket : this.tickets) {
             price += ticket.getPrice();
         }
         return price;
     }
+
+    public void setUser(User user) {
+        this.user = user;
+        if (user != null) {
```
martijnjanssen: We could also have check in here that throws an exception? An `assert(...)` is also possible. This looks quite weird to me, checking null-ness on a setter method.

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
martijnjanssen: That is correct, it is not flushed to the database yet, but when failure occurs, data will not be saved in the database yet and lost. If you read [this](http://stackoverflow.com/questions/21203875/difference-between-save-and-saveandflush-in-spring-data-jpa) stackoverflow answer, the changes will only be written to the database until the flush or commit methods are called. I would think the saveAndFlush method would fit better in our current usage of all the methods. We can change later to a save and commit/flush structure.

martijnjanssen:
Really only finding small issues now, it looks very good!

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
skamoen: Let's create a separate issue for this. It's probably worth investigating, but covers more then just this PR.

src/main/java/ch/wisv/areafiftylan/users/model/UserDTO.java
skamoen: I think this is needed for the auto binding magic. It doesn't work without it.

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
martijnjanssen: I mean that using the usual saveAndFlush would be better untill changing to the other method.

src/main/java/ch/wisv/areafiftylan/products/model/Order.java
skamoen: It's used a lot in the OrderServiceTest. As this is just the model constructor, I don't think it's a problem that we have two constructors. Both work fine and mixing usage doesn't break anything.

src/main/java/ch/wisv/areafiftylan/products/model/Order.java
skamoen: I implemented it like this to allow for "unassigning" an Order. Can't really think of a usecase though

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -187,10 +213,14 @@ public Order updateOrderStatus(Long orderId) {
 
     @Override
     public void adminApproveOrder(Long orderId) {
```
skamoen: Let's leave this for now, and make an issue to replace with the functionality you suggest

martijnjanssen:
Looks really good!

PR #314, opened 2016-12-03
Dynamic Tickets

### LINKED ISSUE #246: Tests fail on date-dependent test
The deadlines for tickets are hardcoded in the Enums. This makes tests fail after a deadline. This should be made dynamic or independent from the tests.

### LINKED ISSUE #262: Dynamic ticket options

skamoen:
This PR introduces a completely new type of Tickets. Tickets can now be added dynamically, instead of relying on a hardcoded Enum.

As this breaks everything, I'm going to need some help finalizing. Use this PR to track progress and mark things as done!

- [x] Create new `X-Auth-Token` abstract IntegrationTest class
- [x] Create new TicketService(Unit)Test
- [x] Everything else...

Fixes #246
Fixes #262


src/main/java/ch/wisv/areafiftylan/exception/TicketOptionNotFoundException.java
```diff
@@ -15,18 +15,10 @@
...
-
-    final float price;
-
-    TicketOptions(float price) {
-        this.price = price;
-    }
-
-    public float getPrice() {
-        return price;
+public class TicketOptionNotFoundException extends RuntimeException {
```
TimvdLippe: Shouldn't this extend `IllegalArgumentException`? Also for the other exceptions down below

src/main/java/ch/wisv/areafiftylan/products/model/TicketType.java
```diff
@@ -17,46 +17,62 @@
...
+    private String text;
+    @Getter
+    @Setter
+    private LocalDateTime deadline;
+    @Getter
+    @Setter
+    private boolean buyable;
+    @OneToMany(cascade = CascadeType.MERGE)
+    @Getter
+    Collection<TicketOption> possibleOptions;
```
TimvdLippe: Shouldn't this be a set? Then you can also use sets in the methods in the service.

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -157,8 +165,12 @@ public Order removeTicketFromOrder(Long orderId, TicketType type, boolean pickup
         }
     }
 
-    private static Predicate<Ticket> isEqualToInput(TicketType type, boolean pickupService, boolean chMember) {
-        return t -> (t.getType() == type) && (t.isChMember() == chMember) && (t.hasPickupService() == pickupService);
+    private static Predicate<Ticket> isEqualToInput(String type, List<String> options) {
+
+        return t -> (t.getType().getName().equals(type) && t.getEnabledOptions().stream().
```
TimvdLippe: I wonder if there is a standard function to check if all options are in a list.

src/main/java/ch/wisv/areafiftylan/products/model/TicketType.java
martijnjanssen: A set would also represent the sort of structure better, ticket types are unique, thus a set would only make this more obvious.

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
skamoen: There's `containsAll()` but I want an exclusive contains. I don't think that exists

TimvdLippe: If you make it a set, you can check on `containsAll` and if the `size` are equal

skamoen:
As it seems impossible to convert gradually, this PR will involve the migration to XAuth Integration Test. As every test needs at least a modifcation, we're also using this to upgrade to Rest Assured 3.0.1. 

Steps for migration:

1. Remove @Before and @After methods, instead create private functions either in `XAuthIntegrationTest` or in the specific `IntegrationTest`. These private functions can work directly in the database to create the initial state. 
2. Reimport `RestAssured` static methods such as `given()` from the new `io.restassured` package
3. Create Users for each test with the helper functions defined in `XAuthIntegrationTest`
4. Look at the currently modified tests for examples. Try to only test one request per test, as this will greatly increase speed.

MatthijsKok:
When updating tests for Rest Assured 3.0.1, please keep in mind the [non backwards compatible changes](https://github.com/rest-assured/rest-assured/wiki/releasenotes30#non-backward-compatible-changes)

src/main/java/ch/wisv/areafiftylan/Application.java
```diff
@@ -42,15 +34,16 @@ public static void main(String[] args) {
...
+    //     */
+    //    @Bean
+    //    @Profile("dev")
+    //    CommandLineRunner init(UserRepository accountRepository, TicketRepository ticketRepository, SeatService
+    // seatService,
+    //                           TeamRepository teamRepository) {
+    //
+    //        return new TestDataRunner(accountRepository, ticketRepository, teamRepository, seatService);
+    //
+    //    }
```
skamoen: This can be removed as TestDataRunner is a `@Component` now

src/main/java/ch/wisv/areafiftylan/exception/TicketOptionNotFoundException.java
skamoen: A 404 is not an Illegal Argument per se.

src/main/java/ch/wisv/areafiftylan/products/service/repository/TicketRepository.java
```diff
@@ -15,19 +15,23 @@
...
 import java.util.Collection;
+import java.util.Optional;
 
 @Repository
 public interface TicketRepository extends JpaRepository<Ticket, Long> {
 
-    Collection<Ticket> findByPickupService_True();
+    Optional<Ticket> findById(Long ticketId);
+
+    //FIXME: Build a new query using options.contains or something
```
skamoen: Fixed as a service call

PR #316, opened 2016-12-14
Refactor TicketRestIntegrationTest to XAuth based authentication

MatthijsKok:
- [ ]  testGetTicketInControlTeamMemberTicket
- [ ]  testGetAllTicketsForTransportAsAdmin
- [ ]  testGetTicketInControlMemberOfTeam
    These are the only tests needed for this class to pass.

### DIFF
--- a/src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
+++ b/src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
@@ -531,7 +531,7 @@ public class TicketRestIntegrationTest extends XAuthIntegrationTest {
     @Test
     public void testGetAllTicketsForTransportAsAdmin() {
         User admin = createUser(true);
-        createTicket(admin, Collections.singletonList("pickupService"));
+        createTicket(admin, Collections.singletonList(PICKUP_SERVICE));
 
         //@formatter:off
         given().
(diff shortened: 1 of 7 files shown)


src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -68,517 +54,493 @@
...
     public void testGetTicketInControlTeamMemberTicket() {
-        User teamMate = createTeamReturnSingleMember();
+        User user = createUser();
+        createTicketForUser(user);
+        User teammate = createUser();
+        createTicketForUser(teammate);
 
-        SessionData login = login(user.getUsername(), userCleartextPassword);
+        Team team = createTeamWithCaptain(user);
+        team.addMember(teammate);
```
skamoen: This is not persisted in the database, which is why the request only has a team with one member.

src/test/java/ch/wisv/areafiftylan/TicketRestIntegrationTest.java
```diff
@@ -68,517 +54,493 @@
...
     @Test
     public void testGetTicketInControlMemberOfTeam() {
-        User teamMate = createTeamReturnSingleMember();
+        User user = createUser();
+        User teammate = createUser();
+        createTicketForUser(teammate);
 
-        SessionData login = login(teamMate.getUsername(), teamMemberCleartextPassword);
+        Team team = createTeamWithCaptain(user);
+        team.addMember(teammate);
```
skamoen: Same here, this change is not persisted

skamoen:
Merging this without thorough test review. We'll do this in #314.

PR #318, opened 2016-12-14
Fix application.properties.sample Thymeleaf configuration

TimvdLippe:
The default sample was incorrect as the Thymeleaf templates were moved to `config/templates`


martijnjanssen:
LGTM!

PR #320, opened 2016-12-22
Add age check endpoint

MatthijsKok:
This checks if the  user is older than a51.user.alcoholage and returns a boolean wether they are or not.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/users/controller/UserRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/controller/UserRestController.java
@@ -149,10 +149,10 @@ public class UserRestController {
         return createResponseEntity(HttpStatus.OK, "User disabled");
     }
 
-    @PreAuthorize("hasRole('ADMIN')")
-    @GetMapping("/{userID}/agecheck")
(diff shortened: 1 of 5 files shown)


src/main/java/ch/wisv/areafiftylan/users/controller/UserRestController.java
```diff
@@ -148,6 +149,15 @@ public User getUserById(@PathVariable Long userId) {
         return createResponseEntity(HttpStatus.OK, "User disabled");
     }
 
+    @PreAuthorize("hasRole('ADMIN')")
```
martijnjanssen: I would change this checking method, both the `ADMIN` and `COMMITTEE` role should have access.

src/main/java/ch/wisv/areafiftylan/users/controller/UserRestController.java
```diff
@@ -148,6 +149,15 @@ public User getUserById(@PathVariable Long userId) {
...
+    @GetMapping("/{userID}/agecheck")
```
martijnjanssen: I assume this is a typo and should be `/{userId}/agecheck`, should it not? It does not match the method argument right now.

PR #321, opened 2016-12-27
Change removeTicket from order to id

martijnjanssen:
Ticket removal is quite weird to remove a random ticket whichever is found first. It is more logical to be able to remove on ticketId in an Order.

Also, tests are failing but I can't figure out what is happening right now. I suppose this is due to the fact that the OrderRestIntegrationTest has not been updated yet? It gives me a failing test, but it does succeed when the test is run with no other tests.


skamoen:
Lets revisit (and rebase) after dynamic tickets? That's partially why the tests are failing too

martijnjanssen:
Looks good!

PR #324, opened 2016-12-30
TicketServiceTest and Unit tests abstract method.

MatthijsKok:


src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    @Test
+    public void getAllTicketsTwo() {
+        persistTicket();
+        persistTicket();
+        Collection<Ticket> tickets = ticketService.getAllTickets();
+        assertEquals(2, tickets.size());
+    }
+
+    @Test
+    public void validateTicketFromInvalid() {
```
martijnjanssen: Test name could be changed to `validateTicket`, this is just the 'base' test. A more thorough test can be given a more specific name.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        Ticket ticket = persistTicket();
+        ticket.setValid(true);
+        testEntityManager.persist(ticket);
+        assertEquals(true, ticket.isValid());
+        ticketService.validateTicket(ticket.getId());
+        assertEquals(true, ticket.isValid());
+    }
+
+    @Test
+    public void validateTicketNotFound() {
```
martijnjanssen: Isn't this just another getter test?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        Long id = 9999L;
+        thrown.expect(TicketNotFoundException.class);
+        thrown.expectMessage("Ticket not found");
+        ticketService.validateTicket(id);
+    }
+
+    @Test
+    public void validateTicketNull() {
```
martijnjanssen: Same for this one.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    @Test
+    public void requestTicketOfTypeStrings() {
+        User user = persistUser();
+        TicketType type = ticketTypeRepository.findByName(TEST_TICKET).get();
+        TicketOption pickupOption = ticketOptionRepository.findByName(PICKUP_SERVICE_OPTION).get();
+        TicketOption chMemberOption = ticketOptionRepository.findByName(CH_MEMBER_OPTION).get();
+        List<TicketOption> options = Arrays.asList(pickupOption, chMemberOption);
+        Set<TicketOption> optionSet = new HashSet<>(options);
```
martijnjanssen: You can insert the previous `Arrays.asList` call in the `HashSet` constructor, removes one line and allows another newline for readability.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+
+    @Test
+    public void validateTicketNull() {
+        thrown.expect(TicketNotFoundException.class);
+        thrown.expectMessage("Ticket not found");
+        ticketService.validateTicket(null);
+    }
+
+    @Test
+    public void requestTicketOfTypeStrings() {
```
martijnjanssen: What is with the strings part in the method name here? It does not really make anything clear or add something. I think it could be removed, but prove me wrong.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        List<String> optionsString = Arrays.asList(CH_MEMBER_OPTION, PICKUP_SERVICE_OPTION);
+
+        thrown.expect(TicketTypeNotFoundException.class);
+        thrown.expectMessage("TicketType ");
+
+        ticketService.requestTicketOfType(user, typeString, optionsString);
+    }
+
+    @Test
+    public void requestTicketOfTypeStringsTypeNull() {
```
martijnjanssen: Does this test really differ from the previous one? What is your opinion on this @skamoen?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        User user = persistUser();
+        TicketType type = ticketTypeRepository.findByName(TEST_TICKET).get();
+        Ticket ticketToPersist = new Ticket(user, type);
+        TicketOption pickupOption = ticketOptionRepository.findByName(PICKUP_SERVICE_OPTION).get();
+        ticketToPersist.addOption(pickupOption);
+        TicketOption chMemberOption = ticketOptionRepository.findByName(CH_MEMBER_OPTION).get();
+        ticketToPersist.addOption(chMemberOption);
+        List<TicketOption> options = Arrays.asList(pickupOption, chMemberOption);
+        Set<TicketOption> optionSet = new HashSet<>(options);
+        testEntityManager.persist(ticketToPersist);
```
martijnjanssen: Why are you persisting a ticket and requesting a new one as the next action? The persist action shouldn't be needed right? This also applies for the other tests.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+
+        Ticket ticket = ticketService.requestTicketOfType(user, type, options);
+
+        assertEquals(user, ticket.getOwner());
+        assertEquals(optionSet, ticket.getEnabledOptions());
+        assertEquals(type, ticket.getType());
+    }
+
+    @Test
+    public void requestTicketOfTypeObjectsTypeNotAvailable() {
```
martijnjanssen: I would change the test name to `requestTicketDeadlinePassed` since that is the behavior you are trying to test, is that correct?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    @Test
+    public void transferTicketTokenNotFound() {
+        thrown.expect(TokenNotFoundException.class);
+        thrown.expectMessage("Could not find token ");
+
+        ticketService.transferTicket("invalid_token");
+    }
+
+    @Test
+    public void transferTicketTokenNull() {
```
martijnjanssen: Isn't this the same as an invalid token?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        ttt.use();
+
+        thrown.expect(InvalidTokenException.class);
+        thrown.expectMessage("Token is expired, has been already used or has been revoked.");
+
+        ticketService.transferTicket(token);
+    }
+
+    @Test
+    public void transferTicketTicketAlreadyLinked() {
```
martijnjanssen: Double `'Ticket'` here? :slightly_smiling_face:

martijnjanssen: This is a good test though, a user should not be able to transfer the ticket after the event has started when the ticket was set up for transfer before checking in.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    @Test
+    public void cancelTicketTransferTokenNotFound() {
+        thrown.expect(TokenNotFoundException.class);
+        thrown.expectMessage("Could not find token ");
+
+        ticketService.cancelTicketTransfer("invalid_token");
+    }
+
+    @Test
+    public void cancelTicketTransferTokenNull() {
```
martijnjanssen: Same as the test above?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        User captain = persistUser();
+        Ticket captainTicket = persistTicketForUser(captain);
+
+        Collection<Ticket> tickets = ticketService.getOwnedTicketsAndFromTeamMembers(captain);
+
+        assertEquals(tickets, Collections.singletonList(captainTicket));
+    }
+
+    @Test
+    public void getOwnedTicketsAndFromTeamMembersNull() {
```
martijnjanssen: This is a test for the userRepository in my opinion and should not be here.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    }
+
+    @Test
+    public void getValidTicketTransferTokensByUserNotFound() {
+        Collection<TicketTransferToken> tttc = ticketService.getValidTicketTransferTokensByUser("doesnt_exist");
+        assertTrue(tttc.isEmpty());
+    }
+
+    @Test
+    public void getValidTicketTransferTokensByUserNull() {
```
martijnjanssen: Same as above?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+    @Test
+    public void getAllTicketsWithTransportZero() {
+        Ticket ticketWithoutPickup =
+                ticketService.requestTicketOfType(TEST_TICKET, Collections.singletonList(CH_MEMBER_OPTION));
+        ticketWithoutPickup.setValid(true);
+        testEntityManager.persist(ticketWithoutPickup);
+
+        Collection<Ticket> tickets = ticketService.getAllTicketsWithTransport();
+
+        assertEquals(tickets, Collections.emptyList());
```
martijnjanssen: A size assert is also sufficient here.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        User user = persistUser();
+        User owner = persistUser();
+        Ticket ticket = persistTicketForUser(owner);
+        assertEquals(owner, ticket.getOwner());
+        Ticket result = ticketService.assignTicketToUser(ticket.getId(), user.getUsername());
+        assertEquals(user, result.getOwner());
+    }
+
+    @Test
+    public void addTicketTypeOne() {
```
martijnjanssen: `addTicketType`?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        long countBefore = ticketTypeRepository.count();
+        TicketType ticketType = new TicketType("type1", "text", 5F, 0, LocalDateTime.now(), true);
+        ticketService.addTicketType(ticketType);
+        assertEquals(countBefore + 1, ticketTypeRepository.count());
+    }
+
+    @Test
+    public void addTicketTypeTwo() {
```
martijnjanssen: `addTwoTicketTypes`?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        long countBefore = ticketTypeRepository.count();
+        TicketType ticketType1 = new TicketType("type1", "text", 5F, 0, LocalDateTime.now(), true);
+        ticketService.addTicketType(ticketType1);
+        TicketType ticketType2 = new TicketType("type2", "text", 6F, 0, LocalDateTime.now(), true);
+        ticketService.addTicketType(ticketType2);
+        assertEquals(countBefore + 2, ticketTypeRepository.count());
+    }
+
+    @Test
+    public void getAllTicketTypesZeroExtra() {
```
martijnjanssen: `getAllTicketTypes`?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        TicketType ticketType1 = new TicketType("type1", "text", 5F, 0, LocalDateTime.now(), true);
+        ticketService.addTicketType(ticketType1);
+        TicketType ticketType2 = new TicketType("type2", "text", 6F, 0, LocalDateTime.now(), true);
+        ticketService.addTicketType(ticketType2);
+        Collection<TicketType> types = ticketService.getAllTicketTypes();
+        assertEquals(countBefore + 2, types.size());
+    }
+
+    @Test
+    public void addTicketOptionOne() {
```
martijnjanssen: `addTicketOption`?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        long countBefore = ticketOptionRepository.count();
+        TicketOption option = new TicketOption("option1", 2.5F);
+        ticketService.addTicketOption(option);
+        assertEquals(countBefore + 1, ticketOptionRepository.count());
+    }
+
+    @Test
+    public void addTicketOptionTwo() {
```
martijnjanssen: `addTwoTicketOptions`?

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
```diff
@@ -282,6 +282,9 @@ private TicketTransferToken getTicketTransferTokenIfValid(String token) {
 
     @Override
     public Collection<Ticket> getOwnedTicketsAndFromTeamMembers(User u) {
+        if (u == null) {
+            throw new UserNotFoundException("null");
```
MatthijsKok: Should we then always assume our services don't receive 'null' values? A lot of methods simply call a JPA method, which in his turn then throws an IllegalDataAccessException. Testing for that would be stupid, and not the output our service is supposed to give. I agree these lines are not correct.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -0,0 +1,664 @@
...
+        thrown.expect(TicketNotFoundException.class);
+        thrown.expectMessage("Ticket not found");
+        ticketService.getTicketById(id);
+    }
+
+    @Test
+    public void getTicketByIdNull() {
+        thrown.expect(TicketNotFoundException.class);
+        thrown.expectMessage("Ticket not found");
+        ticketService.getTicketById(null);
```
MatthijsKok: Services don't deal with HTTP status codes.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
MatthijsKok: there are two "requestTicketOfType()" methods, one that accepts objects, and one that accepts strings.

martijnjanssen: But why the distinction, are we using both?

MatthijsKok: The one that accepts strings attempts to find the objects, and then calls the second method. The second method is not used elsewhere. both are available to the controller, so both should be tested for now. Do you think we need to unify these methods?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
MatthijsKok: Because the requesting is what we test. The test then asserts if the retrieved `ticket` is equal to the one we persisted. How else could we test this method?

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
MatthijsKok: It is. However, in OrderServiceTest, testing of invalid and null is also done, and these tests are heavily inspired by it.

PR #325, opened 2017-01-01
Order integration test

skamoen:


src/test/java/ch/wisv/areafiftylan/OrderRestIntegrationTest.java
```diff
@@ -116,959 +80,544 @@ public void testViewAllOrders_User() {
...
-            body("object.user", is(nullValue())).
-            body("object.status", is("ANONYMOUS")).
-            body("object.tickets", hasSize(1)).
-            body("object.tickets.pickupService", hasItem(true)).
-            body("object.tickets.type", hasItem(is(TicketType.TEST.toString()))).
-            body("object.amount",equalTo(TicketType.TEST.getPrice() + TicketOption.PICKUPSERVICE.getPrice()));
+            get(ORDER_ENDPOINT).
+        then().
+            statusCode(HttpStatus.SC_OK).
+            body("$", hasSize(Long.valueOf(orderRepository.count()).intValue()));
```
martijnjanssen: Do we need to check if expired orders are included in this return method?

skamoen: The archived expired order are in a completely different repo, and the 30 minutes they're expired as an Order, it's not unwanted to have them here.

PR #328, opened 2017-01-02
Order confirmation

### LINKED ISSUE #136: Email confirmation of order
With a successful order, it would be good to email the user with a confirmation.

skamoen:
This PR adds an Orderconfirmation email after an Orderstatus is updated to PAID.

Currently this only works through the updateByReference method, which is used by the webhook of Mollie. Manual updates go through updateById. Not sure what is desired

~Blocked by #314~
Fixes #136


martijnjanssen:
Looks good! We kinda needed this from day one.

PR #330, opened 2017-01-03
Unify updating of OrderStatus

skamoen:
We had two methods, one via OrderID, another one via OrderReference. They should do the same, they didn't.


martijnjanssen:
Nice change, useful for testing purposes!

PR #331, opened 2017-01-04
Implement JsonUsernamePasswordAuthentication via Filter

### LINKED ISSUE #329: When user register, they don't have to confirm their account to log in

skamoen:
With the transition to Auth Tokens, we started using some super simple authentication. We ran into some problems like #329. This PR implements an extension of the old authentication method, with some custom json parsing. Should be more secure then our own fiddling.

Changes for frontend:
Login at `/login` instead of `/token`
Token can be found in the `X-Auth-Token` header, not in the body

Fixes #329


martijnjanssen:
Great stuff! 10/10 would merge again!

PR #332, opened 2017-01-04
Improve consistency in SeatRestController.

### LINKED ISSUE #200: Make Seat reservation more consistent
Seat reservation for `User`s requires the `seatId` as a _parameter_, instead of the body. This is the only place in the API that uses params, and it's used in the wrong way. The admin function that uses a `SeatReservationDTO` should be merged with the `User` function. Admin functionality can be checked in the service.

MatthijsKok:
Tries to improve consistency in SeatRestController.

Fixes #200.
Needs frontend updating, as endpoints are changed.
[Frontend issue](https://github.com/AreaFiftyLAN/lancie-frontend/issues/351)


src/main/java/ch/wisv/areafiftylan/seats/controller/SeatRestController.java
```diff
@@ -205,7 +206,7 @@ SeatmapResponse getAllSeatsAdminView() {
      * @return The Seat of the given User.
      */
     @PreAuthorize("hasRole('ADMIN')")
-    @RequestMapping(value = "/users/{userId}/seat", method = RequestMethod.GET)
+    @GetMapping("/user/{userId}")
```
skamoen: I think this violates REST design

src/main/java/ch/wisv/areafiftylan/seats/controller/SeatRestController.java
```diff
@@ -221,7 +222,7 @@ SeatmapResponse getAllSeatsAdminView() {
      */
     @JsonView(View.Public.class)
     @PreAuthorize("isAuthenticated()")
-    @RequestMapping(value = "/teams/{teamName}/seats", method = RequestMethod.GET)
+    @GetMapping("/team/{teamName}")
     public Collection<Seat> getSeatsForTeam(@PathVariable String teamName) {
```
skamoen: This one too

src/main/java/ch/wisv/areafiftylan/seats/controller/SeatRestController.java
MatthijsKok: I Agree. I think we should move these methods out of the SeatRestController.

PR #333, opened 2017-01-04
Also allow captains of teams containing only them to delete the team.

### LINKED ISSUE #227: Impossible to delete team
Would be nice if you could leave the team even as a last member or just have the option to delete it.

MatthijsKok:
Fixes #227.
Needs frontend button to delete team when `team.members.size == 1`.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
@@ -95,19 +95,6 @@ public class CurrentUserServiceImpl implements CurrentUserService {
     }
 
     @Override
-    public boolean canDeleteTeam(Object principal, Long teamId) {
-        if (principal instanceof UserDetails) {
-            UserDetails user = (UserDetails) principal;
-            Team team = teamService.getTeamById(teamId);
-            return (team.getCaptain().getUsername().equals(user.getUsername()) &&
-                    team.getMembers().size() == 1) ||
-                    user.getAuthorities().contains(Role.ROLE_ADMIN);
-        } else {
-            return false;
-        }
-    }
-
-    @Override
     public boolean canRemoveFromTeam(Object principal, Long teamId, String username) {
         if (principal instanceof UserDetails) {
             UserDetails currentUser = (UserDetails) principal;
--- a/src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
@@ -95,6 +95,19 @@ public class CurrentUserServiceImpl implements CurrentUserService {
     }
 
     @Override
+    public boolean canDeleteTeam(Object principal, Long teamId) {
+        if (principal instanceof UserDetails) {
+            UserDetails user = (UserDetails) principal;
+            Team team = teamService.getTeamById(teamId);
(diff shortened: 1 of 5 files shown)


src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
```diff
@@ -95,6 +95,19 @@ public boolean canEditTeam(Object principal, Long teamId) {
     }
 
     @Override
+    public boolean canDeleteTeam(Object principal, Long teamId) {
```
skamoen: It feels a bit wrong to involve the state of a Team in this service, which should deal with User authorizations. I propose we move this logic to the Controller. It is a matter of authorization (which controllers handle) so I think some controller logic is permitted there. You can use the canEditTeam() PreAuthorize for that method instead.

src/main/java/ch/wisv/areafiftylan/teams/controller/TeamRestController.java
```diff
@@ -232,6 +232,21 @@ public Team update(@PathVariable Long teamId, @Validated @RequestBody TeamDTO in
...
+     * Delete the Team with the given Id. Can only be done by an admin,
+     * or the captain of the team when it has no other members.
+     *
+     * @param teamId Id of the Team to be deleted.
+     * @return A status message of the operation
+     */
+    @PreAuthorize("@currentUserServiceImpl.canDeleteTeam(principal, #teamId)")
+    @JsonView(View.Public.class)
+    @DeleteMapping("/{teamId}")
+    public ResponseEntity<?> deleteTeam(@PathVariable Long teamId) {
```
skamoen: Move the CurrentUserService Logic here. You can return a custom error response when the team is not in the right state.

src/main/java/ch/wisv/areafiftylan/teams/controller/TeamRestController.java
```diff
@@ -232,6 +232,21 @@ public Team update(@PathVariable Long teamId, @Validated @RequestBody TeamDTO in
     }
 
     /**
+     * Delete the Team with the given Id. Can only be done by an admin,
+     * or the captain of the team when it has no other members.
+     *
+     * @param teamId Id of the Team to be deleted.
+     * @return A status message of the operation
+     */
+    @PreAuthorize("@currentUserServiceImpl.canDeleteTeam(principal, #teamId)")
```
skamoen: Use the canEditTeam() here

src/main/java/ch/wisv/areafiftylan/teams/service/TeamServiceImpl.java
```diff
@@ -113,6 +112,10 @@ public Team update(Long teamId, TeamDTO input) {
 
     @Override
     public Team delete(Long teamId) {
+        List<TeamInviteResponse> invites = findTeamInvitesByTeamId(teamId);
```
skamoen: Good catch. But perhaps we can do this automatically with some cascading?

MatthijsKok: I don't know if cascading could work, since the `TeamInviteToken `has the `Team`, so when the `Team` gets deleted it has no knowledge of `TeamInviteToken`s.

martijnjanssen:
Sorry for being so late to the party but quick question: Why did we implement this this way? Couldn't this be handled by deleting the team when the last user is being deleted? This would require one action less.

MatthijsKok:
Uhh sure could've done that. I guess it didn't occur to me then.

PR #341, opened 2017-01-06
Add check if ticket transfer is going to the same user.

skamoen:
That wouldn't make much sense huh. API now returns 400 if this is attempted


MatthijsKok:
Fixes #340

PR #344, opened 2017-01-08
Fix user controllers

MatthijsKok:
Some methods with preceding "/users/current" were not in the CurrenrUserRestController.
Also move to @AuthenticationPrincipal, to improve type safety.


src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
```diff
@@ -185,4 +185,10 @@
     public List<Seat> getCurrentUserSeat(@AuthenticationPrincipal User user) {
         return seatService.getSeatsByUsername(user.getUsername());
     }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @GetMapping("/admin")
+    public ResponseEntity<?> adminCheck() {
```
skamoen: You know `GET /users/current` gives you a list of roles right?

MatthijsKok:
This is fixed in #407

PR #354, opened 2017-01-12
Feature ticket type methods

### LINKED ISSUE #353: Add endpoint to change TicketType.
This is to extend the early bird deadline beyond stufi.

MatthijsKok:
Fixes #353

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
@@ -133,7 +133,11 @@ public class TicketRestController {
     @PutMapping("/types/{typeId}")
     public ResponseEntity<?> updateTicketType(@PathVariable Long typeId, @RequestBody @Validated TicketType type) {
         TicketType ticketType = ticketService.updateTicketType(typeId, type);
-        return createResponseEntity(HttpStatus.OK, "TicketType successfully updated.", ticketType);
+        if (ticketType == null) {
+            return createResponseEntity(HttpStatus.BAD_REQUEST, "TicketType not found!");
+        } else {
+            return createResponseEntity(HttpStatus.OK, "TicketType successfully updated.", ticketType);
+        }
     }
 
     @PreAuthorize("hasRole('ADMIN')")
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
@@ -119,10 +119,29 @@ public class TicketRestController {
     @PreAuthorize("hasRole('ADMIN')")
     @PostMapping("/types")
     public ResponseEntity<?> addTicketType(@RequestBody @Validated TicketType type) {
-
         TicketType ticketType = ticketService.addTicketType(type);
-
         return createResponseEntity(HttpStatus.CREATED, "TicketType successfully added", ticketType);
+    }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @GetMapping("/types")
+    public ResponseEntity<?> readTicketTypes() {
+        Collection<TicketType> ticketTypes = ticketService.getAllTicketTypes();
+        return createResponseEntity(HttpStatus.OK, "TicketTypes successfully retrieved.", ticketTypes);
+    }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @PutMapping("/types")
+    public ResponseEntity<?> updateTicketType(@RequestBody @Validated TicketType type) {
+        TicketType ticketType = ticketService.updateTicketType(type);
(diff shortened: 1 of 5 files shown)


src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -119,13 +119,32 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
...
         TicketType ticketType = ticketService.addTicketType(type);
-
         return createResponseEntity(HttpStatus.CREATED, "TicketType successfully added", ticketType);
     }
 
     @PreAuthorize("hasRole('ADMIN')")
+    @GetMapping("/types")
+    public ResponseEntity<?> readTicketTypes() {
+        Collection<TicketType> ticketTypes = ticketService.getAllTicketTypes();
+        return createResponseEntity(HttpStatus.OK, "TicketTypes successfully retrieved.", ticketTypes);
```
skamoen: You're getting a collection here, so just return those. No need for a wrapping response entity

src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -119,13 +119,32 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
...
+    }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @PutMapping("/types")
```
skamoen: This mapping implies replacing all TicketTypes

src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -119,13 +119,32 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
...
+    public ResponseEntity<?> updateTicketType(@RequestBody @Validated TicketType type) {
+        TicketType ticketType = ticketService.updateTicketType(type);
```
skamoen: Which type are you actually updating here?

src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -119,13 +119,32 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
...
+        return createResponseEntity(HttpStatus.OK, "TicketType successfully updated.", ticketType);
+    }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @DeleteMapping("/types")
```
skamoen: This implies deleting all TicketTypes. It should be a DELETE on `/types/{id}` or something like that.

src/test/java/ch/wisv/areafiftylan/products/service/TicketServiceTest.java
```diff
@@ -650,6 +652,62 @@ public void getAllTicketTypesTwoExtra() {
...
+        ticketTypeRepository.save(ticketType2);
+        ticketType = ticketService.updateTicketType(ticketType.getId(), ticketType);
+        assertTrue(ticketType.isBuyable());
+    }
+
+    @Test
+    public void updateTicketTypeTestNoneThere() {
+        TicketType ticketType = new TicketType("type1", "text", 5F, 0, LocalDateTime.now(), false);
+        ticketType = ticketService.updateTicketType(ticketType.getId(), ticketType);
+        assertFalse(ticketType.isBuyable());
```
skamoen: I dont think this is the kind of assertion you want for this test

src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -119,13 +119,35 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
...
+    @GetMapping("/types")
+    public Collection<TicketType> readTicketTypes() {
+        return ticketService.getAllTicketTypes();
+    }
+
+    @PreAuthorize("hasRole('ADMIN')")
+    @PutMapping("/types/{typeId}")
+    public ResponseEntity<?> updateTicketType(@PathVariable Long typeId, @RequestBody @Validated TicketType type) {
+        TicketType ticketType = ticketService.updateTicketType(typeId, type);
+        if (ticketType == null) {
```
skamoen: This is a weird check for a controller. When would this be null exactly? You already made the call to the service here...

MatthijsKok: The service returns null if the save was not done successfully (when typeId doesn't exist for example)

skamoen: Then I'd suggest to throw some exceptions in the service for errors like that.

MatthijsKok: Good point. Will do

PR #357, opened 2017-01-16
Add endpoint for retrieval of PaymentURL for a pending order

### LINKED ISSUE #356: Request payment URL for pending orders
For orders with a status where there hasn't been paid yet, it should be possible to request the payment URL again.

skamoen:
This PR adds an endpoint to get the payment URL for pending orders. This can be used to continue an order that was previously started, but interrupted. Only works for Orders that already started (=PENDING). Other others either need to be checked out first, or are already cancelled. 

Fixes #356


MatthijsKok:
Great clean looking code, tested the endpoint and working. Can we add some integration tests that don't succeed the flow?

PR #364, opened 2017-02-02
Web CMS Remove old code. This change is breaking.

MatthijsKok:
Adding back the code piece by piece in seperate PR's to keep them small(ish).


src/main/java/ch/wisv/areafiftylan/web/controller/WebPropertiesController.java
```diff
@@ -1,40 +0,0 @@
...
-import static ch.wisv.areafiftylan.utils.ResponseEntityBuilder.createResponseEntity;
-
-@RestController
-@RequestMapping("/web/properties")
-public class WebPropertiesController {
-
-    @Value("${a5l.googleMapsAPIkey:API_KEY}")
-    private String googleMapsKey;
-
-    @RequestMapping(value = "/googlemapskey", method = RequestMethod.GET)
```
skamoen: We need this, and is currently being used in the front-end

MatthijsKok: This is added back in a separate PR.

PR #371, opened 2017-02-16
Attempt to solve expiration of successful orders

skamoen:
We've had two glitches in the Matrix where users successfully paid for their order, but the following webhook resulted in a 404, as no Order could be found with the indicated reference. The cleanup task happily cleaned up these orders as they were never marked as PENDING. 

If the paymentURL was returned, the Order reference must have been saved. I couldn't find a bug in the code, but these changes make the process a little more robust and transparent in terms of logging. 

Can't reproduce and not sure if this actually fixes it.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
@@ -42,8 +42,6 @@ public class MolliePaymentService implements PaymentService {
 
     private final OrderRepository orderRepository;
 
-    private final Client mollie;
-
     @Value("${a5l.molliekey:null}")
     String apiKey;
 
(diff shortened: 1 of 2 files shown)


src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
```diff
@@ -42,30 +42,27 @@
...
     @Autowired
-    public MolliePaymentService(OrderRepository orderRepository, @Value("${a5l.molliekey:null}") String apiKey) {
+    public MolliePaymentService(OrderRepository orderRepository) {
         this.orderRepository = orderRepository;
-        this.mollie = new ClientBuilder().withApiKey(apiKey).build();
     }
 
     @Override
     public String registerOrder(Order order) {
+        Client mollie =  new ClientBuilder().withApiKey(apiKey).build();
```
martijnjanssen: Could you explain what this call does?

skamoen: This creates a new instance of the Mollie client for each request. Just to rule out any threading issues, as there's a significant delay in communication with Mollie.

skamoen:
No. As I can't determine the exact cause of the errors, there's no guarantee this will fix it. This makes the whole process a little more transparent and robust, but it's just a guess.

PR #372, opened 2017-02-20
Continue filter chain after login

skamoen:
We had some CORS errors, which was because the filter chain was interrupted on successful authentication. This fix allows the filter chain to continue, so the request passes the CORSFilter, adding some required headers.


MatthijsKok:
LGTM

PR #374, opened 2017-02-21
Enable logging in dev

martijnjanssen:
Logging was broken during the update to spring boot 1.5.1 since logging was not enabled during testing. The entire application is now broken and will not start anymore because of this.


martijnjanssen:
Nice, thanks! I assume this is now ready for a merge?

PR #376, opened 2017-02-21
Add Auth header to allowed headers

skamoen:
Fixes the `Refused to get unsafe header "X-Auth-Token"` error for cross origin requests.


martijnjanssen:
I think we also need the [Access-Control-Request-Headers](http://stackoverflow.com/questions/32500073/request-header-field-access-control-allow-headers-is-not-allowed-by-itself-in-pr) since we are now getting a `XMLHttpRequest cannot load https://api.areafiftylan.nl/api/v1/users/current. Request header field x-auth-token is not allowed by Access-Control-Allow-Headers in preflight response`.

martijnjanssen:
Nice work!

PR #378, opened 2017-02-24
Correctly address our users in emails.

MatthijsKok:


martijnjanssen:
Looks good to me, haven't tested it yet though.

src/main/java/ch/wisv/areafiftylan/utils/mail/MailServiceImpl.java
```diff
@@ -137,7 +137,7 @@ public void sendVerificationmail(User user, String url) {
                 "Please click on the following link to complete your registration: <a href=\"" + url + "\">" + url +
                         "</a><br /><br />If the link does not work, please copy the link and" +
                         " paste it into your browser.";
-        sendMail(user.getUsername(), user.getUsername(), "Confirm your registration", message);
+        sendMail(user.getUsername(), user.getProfile().getFirstName() + " " + user.getProfile().getLastName(), "Confirm your registration", message);
```
martijnjanssen: Like Sille said: Please note that this is not possible, the first- and lastname are not set here.

martijnjanssen:
A helper method would be good to check whether the first and lastname are filled in, if not, use the username as name in the email.

skamoen:
LGTM. You can rebase and merge :+1:

PR #379, opened 2017-02-24
Fix pickup service endpoint.

MatthijsKok:
this fixes the broken `/tickets/transport` endpoint, but we should find a non-hardcoded solution.


skamoen:
Dynamic solution is easy. Try harder

skamoen:
Let's do this properly, including export functionality for lancie-admin. Closing this.

PR #393, opened 2017-03-22
Massive lombok usage improvements.

MatthijsKok:


src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
```diff
@@ -22,57 +22,46 @@
...
-    @Setter
     @JsonView(View.Public.class)
-    public boolean taken;
+    private boolean taken = false;
 
-    @Getter
     @OneToOne(cascade = CascadeType.MERGE)
     @JsonView(View.Public.class)
-    public Ticket ticket;
+    private Ticket ticket;
```
skamoen: Did you test if this change affects the JsonView?

src/main/java/ch/wisv/areafiftylan/users/model/User.java
```diff
@@ -35,67 +33,48 @@
...
     @Id
-    @Getter
     private Long id;
 
     @JsonIgnore
     @ElementCollection(targetClass = Role.class, fetch = FetchType.EAGER)
     @Enumerated(EnumType.STRING)
     @CollectionTable(name = "user_role")
-    private Set<Role> roles;
+    private Set<Role> roles = new HashSet<>(Sets.newHashSet(Role.ROLE_USER));
```
skamoen: While we're changing this, I'm not sure this is the best place to assign default Roles. Perhaps move this to the Service?

src/main/java/ch/wisv/areafiftylan/products/model/TicketOption.java
```diff
@@ -19,39 +19,28 @@
...
+    @NonNull
     @JsonView(View.OrderOverview.class)
     private String name;
 
-
-    public TicketOption(String name, float price) {
-        this.name = name;
-        this.price = price;
-    }
+    @NonNull
```
MatthijsKok: https://projectlombok.org/features/Constructor.html
Lombok's NonNull injects a null check, and marks that field to get a parameter in the @RequiredArgsConstructor.
The latter is why it's used here.

src/main/java/ch/wisv/areafiftylan/products/model/TicketOption.java
```diff
@@ -19,39 +19,28 @@
...
     private Long id;
 
-    @Getter
-    @Setter
-    @JsonView(View.OrderOverview.class)
-    private float price;
-
-    @Getter
-    @Setter
+    @NonNull
```
MatthijsKok: https://projectlombok.org/features/Constructor.html
Lombok's NonNull injects a null check, and marks that field to get a parameter in the @RequiredArgsConstructor.
The latter is why it's used here.

src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
MatthijsKok: No i have not. Will do and report back.

src/main/java/ch/wisv/areafiftylan/products/model/TicketOption.java
skamoen: Point remains. a `float` will never be `null`, it defaults to 0.

MatthijsKok:
If https://github.com/rzwitserloot/lombok/issues/1376 ever gets implemented, this could be done much better

rspilker:
Matthijs, can you contact me a roel@projectlombok.org?

src/main/java/ch/wisv/areafiftylan/users/model/User.java
```diff
@@ -75,6 +77,19 @@
     @JsonIgnore
     private boolean enabled = true;
 
+    public User(String email, String passwordHash) {
+        this.email = email;
+        this.passwordHash = passwordHash;
+        this.profile = new Profile();
+        this.roles = new HashSet<>();
```
TimvdLippe: Previously you initialized it with `Sets.newHashSet(Role.ROLE_USER)`, intended to be removed? Also you already initialized this field, so I don't think again initializing is required

MatthijsKok: Correct. Now the default roles are assigned in the Service

src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
MatthijsKok: I tested, does not affect the JsonView

src/main/java/ch/wisv/areafiftylan/extras/rfid/model/RFIDLink.java
```diff
@@ -19,23 +19,30 @@
...
+@Data
+@NoArgsConstructor
 public class RFIDLink {
+
     @Id
     @GeneratedValue
     private Long id;
 
+    @NonNull
+    @Length(min = RFID_CHAR_COUNT, max = RFID_CHAR_COUNT)
```
skamoen: As this is a Hibernate validator, it only gets checked if we call `@Validated` in a Controller somewhere. Is this the case? I think we only do that for the DTO. Anyway, is this covered by a test to make sure we're not getting a false sense of security here?

MatthijsKok: You're right, we don't. The current tests we have are lacking and strange. I will remove this and handle it in a seperate Issue.

PR #394, opened 2017-04-17
Add actuator endpoints at /management

### LINKED ISSUE #238: Rework custom exceptions
Exceptions leave a huge stack-trace in the logs, while this is not always neccesary. We should make a generic AreaFiftyLANException which does this: http://stackoverflow.com/questions/31250598/prevent-stack-trace-logging-for-custom-exception-in-spring-boot-application

```
public class AreaFiftyLANException extends RuntimeException {
    @Override
    public synchronized Throwable fillInStackTrace() {
        return this;
    }
}
```

And extend our custom exceptions on this. All exceptions thrown should be checked for clarity, as this would eliminate the stracktrace.

skamoen:
This adds Spring Boot Actuator endpoints on `/management`. See http://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html

Pretty useful, but not all of them. This was initially added for the `/management/logfile` endpoint. 

Todo:

- [x] Disable unneccesary endpoints
- [x] Edit log4j2 config to create an additional json logfile
- [ ] Fix #238 
- [x] Write some authentication tests for the management endpoints. We don't want these exposed on accident.


config/application.properties.sample
```diff
@@ -64,6 +64,11 @@ server.context-path=/api/v1
 
 #Log settings for production. Change to log4j2.xml for dev
 logging.config=classpath:log4j2-prod.xml
+logging.file=logs/log.1
+
+#Spring Boot Actuator endpoint settings. Security through SecurityConfiguration.java
+management.security.enabled=false
```
praseodym: why not set `management.security.role=ADMIN` instead?

skamoen: Because that results in a basic-auth prompt, which didn't work for me when I filled in the admin test user.

praseodym:
Also, the `health` endpoint is very useful to have.

config/application.properties.sample
praseodym: Actuator uses Spring Security and `management.security.role=ADMIN` configures it to use the `ADMIN` instead of `ACTUATOR` role. Maybe try pre-authenticating w/o basic auth? If that fails something might be broken in your security config.

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
```diff
@@ -45,9 +44,11 @@
 import static ch.wisv.areafiftylan.utils.ResponseEntityBuilder.createResponseEntity;
 
 @RestController
-@Log4j2
+@Slf4j
 public class OrderRestController {
 
+//    private static final org.slf4j.Logger log = org.slf4j.LoggerFactory.getLogger(OrderRestController.class);
```
martijnjanssen: This can be removed right?

src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
```diff
@@ -208,22 +201,21 @@ public Order getOrderById(@PathVariable Long orderId) {
      * reference
      *
      * @param orderReference Id of the order at the paymentprovider, stored in the reference field
-     *
      * @return Statusmessage
      */
     @RequestMapping(value = "/orders/status", method = RequestMethod.POST)
     public ResponseEntity<?> updateOrderStatus(@RequestParam(name = "id") String orderReference) {
-        log.log(Level.getLevel("A5L"), "Incoming paymentprovicer webhook for reference: {}", orderReference);
+//        log.log(Level.getLevel("A5L"), "Incoming paymentprovicer webhook for reference: {}", orderReference);
```
martijnjanssen: Can also be removed.

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationController.java
```diff
@@ -96,21 +95,20 @@ public AuthenticationController(UserService userService, AuthenticationService a
...
-     *
      * @return A status message telling whether the action was successful
      */
     @RequestMapping(value = "/requestResetPassword", method = RequestMethod.POST)
     @ResponseBody
     public ResponseEntity<?> requestResetPassword(HttpServletRequest request, @RequestBody Map<String, String> body) {
         String username = body.get("username");
 
-        log.log(Level.getLevel("A5L"), "Requesting password reset on email {}.", username);
+//        log.log(Level.getLevel("A5L"), "Requesting password reset on email {}.", username);
```
martijnjanssen: Can also be removed.

skamoen:
We'll be switching to a custom logging service which saves stuff to the database.

praseodym:
Doesn't make the health endpoint any less useful.

PR #398, opened 2017-04-29
Rework custom exceptions

### LINKED ISSUE #238: Rework custom exceptions

MatthijsKok:
Fixes #238


skamoen:
As we're removing the stacktrace, exceptions coming from wrapper functions like `getUserById()` will be useless. Let's review all thrown exceptions first before merging this.

src/main/java/ch/wisv/areafiftylan/exception/AreaFiftyLANException.java
```diff
@@ -0,0 +1,40 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class AreaFiftyLANException extends RuntimeException {
+
+    public AreaFiftyLANException(AreaFiftyLANExceptionLogLevelEnum logLevelEnum, String message) {
+        super(message);
+        switch (logLevelEnum) {
+            case ERROR:
```
TimvdLippe: Instead of a switch, cant you add the log functions in the enum itself?

src/main/java/ch/wisv/areafiftylan/exception/AreaFiftyLANExceptionLogLevelEnum.java
```diff
@@ -0,0 +1,9 @@
+package ch.wisv.areafiftylan.exception;
+
+enum AreaFiftyLANExceptionLogLevelEnum {
+    ERROR,
```
TimvdLippe: E.g. add a constructor here which is a consumer of a string. Then you pass the log reference on the creation of the enum value

PR #406, opened 2017-05-07
Change username to email

martijnjanssen:
All occurrences of username have been changed to email with the exception of some methods regarding logging in, in these cases the username keyword is critical.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
@@ -31,7 +31,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.security.core.Authentication;
-import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 
@@ -81,7 +80,7 @@ public class TicketRestController {
     public ResponseEntity<?> getTicketTokensOpenForTransfer(Authentication auth) {
         User currentUser = (User) auth.getPrincipal();
         Collection<TicketTransferToken> tokens =
-                ticketService.getValidTicketTransferTokensByUser(currentUser.getEmail());
+                ticketService.getValidTicketTransferTokensByUserEmail(currentUser.getEmail());
 
         return createResponseEntity(HttpStatus.OK, "Ticket transfer tokens successfully retrieved.", tokens);
     }
--- a/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
@@ -53,9 +53,9 @@ public class TicketRestController {
 
     @PreAuthorize("@currentUserServiceImpl.isTicketOwner(principal, #ticketId)")
     @RequestMapping(value = "/transfer/{ticketId}", method = RequestMethod.POST)
-    public ResponseEntity<?> requestTicketTransfer(@PathVariable Long ticketId, @RequestBody String goalUsername,
+    public ResponseEntity<?> requestTicketTransfer(@PathVariable Long ticketId, @RequestBody String goalEmail,
                                                    Authentication auth) {
-        TicketTransferToken ttt = ticketService.setupForTransfer(ticketId, goalUsername);
+        TicketTransferToken ttt = ticketService.setupForTransfer(ticketId, goalEmail);
 
         return createResponseEntity(HttpStatus.OK, "Ticket successfully set up for transfer.", ttt.getToken());
     }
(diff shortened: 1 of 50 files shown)


src/main/java/ch/wisv/areafiftylan/products/controller/TicketRestController.java
```diff
@@ -79,9 +79,9 @@ public TicketRestController(TicketService ticketService, OrderService orderServi
     @PreAuthorize("isAuthenticated()")
     @RequestMapping(value = "/tokens", method = RequestMethod.GET)
     public ResponseEntity<?> getTicketTokensOpenForTransfer(Authentication auth) {
-        UserDetails currentUser = (UserDetails) auth.getPrincipal();
+        User currentUser = (User) auth.getPrincipal();
         Collection<TicketTransferToken> tokens =
-                ticketService.getValidTicketTransferTokensByUser(currentUser.getUsername());
+                ticketService.getValidTicketTransferTokensByUser(currentUser.getEmail());
```
skamoen: Method name is wrong here (was, still is). Should either accept a `User`, or be changed to `getValidTicketTransferTokensByUserEmail()`

src/main/java/ch/wisv/areafiftylan/products/service/TicketService.java
```diff
@@ -66,9 +66,9 @@ default Ticket requestTicketOfType(String type, List<String> options) {
      * Sets up the ticket for transfer
      *
      * @param ticketId     The ID of the ticket to be transferred
-     * @param goalUserName The name of the user which should receive the ticket
+     * @param goalEmail    The name of the user which should receive the ticket
      */
-    TicketTransferToken setupForTransfer(Long ticketId, String goalUserName);
+    TicketTransferToken setupForTransfer(Long ticketId, String goalEmail);
```
skamoen: While we're changing names, doesn't `String receiverEmail` make more sense here?

src/main/java/ch/wisv/areafiftylan/products/service/TicketServiceImpl.java
```diff
@@ -183,7 +183,7 @@ public TicketTransferToken setupForTransfer(Long ticketId, String goalUserName)
             throw new TicketAlreadyLinkedException();
         }
 
-        if (t.getOwner() != null && t.getOwner().getUsername().equals(goalUserName)) {
+        if (t.getOwner() != null && t.getOwner().getEmail().equals(goalEmail)) {
```
skamoen: Same here, `receiverEmail`

src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
```diff
@@ -107,15 +106,14 @@
     @JsonView(View.Team.class)
     @RequestMapping(value = "/teams", method = RequestMethod.GET)
     public Collection<Team> getCurrentTeams(Authentication auth) {
-        UserDetails currentUser = (UserDetails) auth.getPrincipal();
-        return teamService.getTeamsByUsername(currentUser.getUsername());
+        User currentUser = (User) auth.getPrincipal();
```
skamoen: I think you can plug in the User directly as a method argument. @MatthijsKok figured this out somewhere. Maybe we can even make it a Class field as every method in this Controller needs the User object.

src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
```diff
@@ -107,15 +106,14 @@
...
+        return teamService.getTeamsByEmail(currentUser.getEmail());
```
skamoen: Let's rename this method to `getTeamsByMemberEmail`

src/main/java/ch/wisv/areafiftylan/utils/db/migration/V20170507210556__rename_username_to_email.java
```diff
@@ -0,0 +1,11 @@
+package ch.wisv.areafiftylan.utils.db.migration;
+
+import org.flywaydb.core.api.migration.spring.SpringJdbcMigration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class V20170507210556__rename_username_to_email implements SpringJdbcMigration {
```
skamoen: Aren't you forgetting a lot of other tables like Orders and ExpiredOrders?

src/test/java/ch/wisv/areafiftylan/UserRestIntegrationTest.java
```diff
@@ -249,23 +249,7 @@ public void testCreateUserIsDisabled() {
     }
 
     @Test
-    public void createUserNoEmail() {
```
skamoen: Why is this removed?

src/main/java/ch/wisv/areafiftylan/utils/db/migration/V20170507210556__rename_username_to_email.java
martijnjanssen: Let me check, but as far as I remember these tables operate with the `Id` in the `Users` table and not the `Username` or `Email`.

src/test/java/ch/wisv/areafiftylan/UserRestIntegrationTest.java
martijnjanssen: Because we already have a [`createUserMissingEmailField()`](https://github.com/AreaFiftyLAN/lancie-api/pull/406/files#diff-1a637b6c2fc39b5c44162281c7a849c2R252) test. It was left over from the move from `username` + `email` to only `email` as username.

src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
martijnjanssen: We'll fix this in @MatthijsKok's PR #344.

PR #409, opened 2017-05-11
Update mappings everywhere.

MatthijsKok:


src/main/java/ch/wisv/areafiftylan/products/controller/OrderRestController.java
```diff
@@ -177,7 +178,7 @@ public Order getOrderById(@PathVariable Long orderId) {
      * @return The paymentURL from the paymentprovider
      */
     @PreAuthorize("@currentUserServiceImpl.canAccessOrder(principal, #orderId)")
-    @GetMapping(value = "/orders/{orderId}/url")
+    @GetMapping(value = "/{orderId}/url")
```
skamoen: You van remove 'value=' here as well

PR #411, opened 2017-05-11
Allow admins to override seats with a new ticket.

### LINKED ISSUE #382: Seat override doesn't work
Currently the SeatServiceImpl checks if the seat is taken and cuts off the procedure.

MatthijsKok:
Fixes #382


MatthijsKok:
Now it is

PR #424, opened 2017-07-22
Move reserveSeat Service logic into a single method. Improve controller

### LINKED ISSUE #228: Admin ticket transfer
Admins should be able to to change the owner of a ticket

### LINKED ISSUE #383: Lock the seatmap except for admins.
This would allow us to make changes without risk of user interference.

### LINKED ISSUE #423: A non-admin user can set any amount of seats as taken when supplying null as ticketId
Will try to add test case to showcase this behaviour.
Related to #228

MatthijsKok:
Also kind of curious what codecov will say.
Fixes #228 
Fixes #383
Fixes #423

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
+++ b/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
@@ -39,7 +39,7 @@ public class Seat {
     private Long Id;
 
     @JsonView(View.Public.class)
-    private boolean taken = false;
+    private boolean locked = true;
 
     @OneToOne(cascade = CascadeType.MERGE)
     @JsonView(View.Public.class)
@@ -62,8 +62,7 @@ public class Seat {
         return ticket.getOwner();
     }
 
-    public void setTicket(Ticket ticket) {
-        this.ticket = ticket;
-        this.taken = this.ticket != null;
+    public boolean isTaken() {
+        return ticket != null;
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
@@ -52,6 +52,13 @@ public class SeatServiceImpl implements SeatService {
     }
 
     @Override
+    public SeatmapResponse getAllSeats() {
+        return new SeatmapResponse(seatRepository.findAll().
+                stream().
+                collect(Collectors.groupingBy(Seat::getSeatGroup)));
+    }
+
+    @Override
     public List<Seat> getSeatsByEmail(String email) {
         return seatRepository.findByTicketOwnerEmailIgnoreCase(email);
 
@@ -69,15 +76,6 @@ public class SeatServiceImpl implements SeatService {
     }
 
     @Override
-    public SeatmapResponse getAllSeats() {
-        List<Seat> all = seatRepository.findAll();
-
-        Map<String, List<Seat>> seatMapGroups = all.stream().collect(Collectors.groupingBy(Seat::getSeatGroup));
-
-        return new SeatmapResponse(seatMapGroups);
-    }
-
-    @Override
     public SeatmapResponse getSeatGroupByName(String groupName) {
         List<Seat> seatGroup = seatRepository.findBySeatGroup(groupName);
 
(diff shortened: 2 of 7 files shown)


MatthijsKok:
I'm thinking about changing the logic of seats though.. We now have a 'taken' flag that is determined by a ticket being set, or an admin taking without a ticket. 
My proposal would be a 'locked' flag that is independent of a ticket being set or not. This also allows us to implement #383 more easily. We would of course still have 'isTaken()' and maybe 'isAvailable() {!isTaken() && !isLocked()}'

MatthijsKok:
I have changed it that all seats that are created are locked by default. It's nice for an admin to be able to take his time adding all seatgroups, and unlock all at once.

src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
```diff
@@ -39,7 +39,7 @@
     private Long Id;
 
     @JsonView(View.Public.class)
-    private boolean taken = false;
```
martijnjanssen: Please check whether we use the taken property in the frontend, changes may be needed there.

src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
```diff
@@ -52,80 +52,73 @@ public SeatServiceImpl(SeatRepository seatRepository, TeamService teamService, T
...
-     * @param seat The Seat to reserve.
-     * @param ticketId The Id of the Ticket to be set to the Seat.
-     */
     @Synchronized
-    private void reserveSeat(Seat seat, Long ticketId) {
+    public boolean reserveSeat(String groupName, int seatNumber, Long ticketId, boolean allowSeatOverride) {
+        Seat seat = getSeatBySeatGroupAndSeatNumber(groupName, seatNumber);
+        Ticket ticket = null;
+
+        if (!allowSeatOverride && (seat.isTaken() || seat.isLocked() || ticketId == null)) {
```
martijnjanssen: Shouldn't it be `ticketId != null` since we are checking for not allowed seat modifications?

src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
```diff
@@ -52,80 +52,73 @@ public SeatServiceImpl(SeatRepository seatRepository, TeamService teamService, T
...
         seatRepository.saveAndFlush(seat);
+        return true;
     }
 
     @Override
-    public Seat getSeatBySeatGroupAndSeatNumber(String groupName, int seatNumber) {
-        return seatRepository.findBySeatGroupAndSeatNumber(groupName, seatNumber)
-                .orElseThrow(SeatNotFoundException::new);
+    public void clearSeat(String groupName, int seatNumber) {
+        reserveSeat(groupName, seatNumber, null, true);
```
martijnjanssen: Shouldn't we always send an email when the seat is cleared, or are you sure that there is no case a seat is cleared and the user is not notified?

src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
MatthijsKok: No. If you do not `allowSeatOverride`, then providing no ticketId is illegal, the same for the seat being taken or locked.

src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
MatthijsKok: Correct. There is now only one method that modifies seats (reserveSeat()), and that has an email check built in.

src/main/java/ch/wisv/areafiftylan/seats/service/SeatServiceImpl.java
martijnjanssen: Ah wait, this is for the `ticketId` I confused it with the `ticketId` on the seat :+1: 

*This is not the comment you are looking for, move along.*

src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
MatthijsKok: We do use that property in the front end, I have created an issue there detailing the needed changes.

PR #426, opened 2017-07-31
Implement Role Hierarchy System

### LINKED ISSUE #387: OPERATOR Role endpoint rules
We would call this Role OPERATOR, this access level is between the level of USER and ADMIN. This would imply that any method requiring OPERATOR would also be accessible to ADMINs.
We should at the same time look into chaining Role levels, like View allows.

MatthijsKok:
I have also tested this by only requiring the Role.OPERATOR on the ConsumptionTest's, and all tests run as admin completed successfully.

The SecurityConfig is now a public class, unsure if any security issues might arise from that.

Fixes #387


skamoen:
Awesome. Next steps are described in #428 and #428

PR #429, opened 2017-07-31
Remove unused HttpServletRequest.

MatthijsKok:


skamoen:
Not sure where this went wrong, but this was there to use the URL where the user came from to send an email containing a link to that same URL. Not sure if we want to remove or fix that

MatthijsKok:
If you check the git history of that method, that functionality was removed in favor of our URLs being in config files

skamoen:
Makes sense, certainly now that we have dev-properties doing just this. Will merge after checks.

PR #432, opened 2017-07-31
Add test to show you can receive a ticket without having a profile

### LINKED ISSUE #253: Can't be a receiver of a ticket transfer if no profile is on the account
Question is whether to give error upon sending or accepting

MatthijsKok:
Fixes #253


skamoen:
#253 is meant to state that this shouldn't be possible, as this becomes a problem at check-in. I suggest we use this PR to actually fix it :)

martijnjanssen:
Yeah, could you fix this @MatthijsKok? :angel:

PR #435, opened 2017-08-17
Additional helper methods for Roles

MatthijsKok:


src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
```diff
@@ -128,49 +146,45 @@ public boolean canAccessOrder(Object principal, Long orderId) {
...
-            return userTeams.stream().map(Team::getMembers).anyMatch(members -> members.contains(ticket.getOwner()));
-        } else {
-            return false;
+            return userTeams.stream().map(Team::getMembers).anyMatch(members -> members.contains(owner));
         }
+        return false;
     }
 
     @Override
     public boolean hasAnyTicket(Object principal) {
```
skamoen: Since we're not restricting the creation of teams to ticketholders, this is no longer used

PR #440, opened 2017-09-15
Correct captain removal

martijnjanssen:
The delete team method was not useful or logical. By allowing this, the flow is much more logical.


src/main/java/ch/wisv/areafiftylan/security/authentication/CurrentUserServiceImpl.java
```diff
@@ -119,11 +119,6 @@ public boolean canRemoveFromTeam(Object principal, Long teamId, String email) {
             User user = (User) principal;
             Team team = teamService.getTeamById(teamId);
 
-            // You can't remove a Team Captain
-            if (team.getCaptain().getEmail().equals(email)) {
```
MatthijsKok: We are removing this since it is not the job of the access control methods.

PR #445, opened 2017-10-23
Subscription model

Fastjur:
For https://github.com/AreaFiftyLAN/lancie-frontend/pull/521

Last thing todo for subscription model. Get a clean error message for unique constraint.
image

## Todo
- [x] Add tests
- [x] Test the remove/unsubscribe functionality (Perhaps this should be done using an id after all, as we'd like this to be able to be a clickable `unsubscribe` link as per CAN-SPAM act)


src/main/java/ch/wisv/areafiftylan/extras/mailupdates/controller/SubscriptionController.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Collection;
+
+import javax.websocket.server.PathParam;
```
martijnjanssen: This should not be here, maybe an `Ctrl+Alt+o` to clean up the imports?

Fastjur: that's weird...

src/main/java/ch/wisv/areafiftylan/extras/mailupdates/service/SubscriptionServiceImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public Collection<Subscription> getSubscriptions() {
+    return subscriptionRepository.findAll();
+  }
+
+  @Override
+  public Subscription getSubscriptionByEmail(String email) {
+    Optional<Subscription> subscription = subscriptionRepository.findByEmail(email);
+    if (!subscription.isPresent()) {
```
martijnjanssen: I think we can use an `orElseThrow` here, which does exactly what you are trying to do here. [Check it out.](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseThrow-java.util.function.Supplier-)

Fastjur: I tried that but it wouldn't work. It was the first thing I wrote, but I got some funky error messages. I am too tired now to fix it though, I'll probably find the issue tomorrow.

src/main/java/ch/wisv/areafiftylan/extras/mailupdates/controller/SubscriptionController.java
```diff
@@ -0,0 +1,86 @@
...
+    this.subscriptionService = subscriptionService;
+  }
+
+  @GetMapping
+  @PreAuthorize("hasRole('COMMITTEE')")
+  public Collection<Subscription> getAllSubscriptions() {
+    return subscriptionService.getSubscriptions();
+  }
+
+  @GetMapping("/{id}")
```
martijnjanssen: Why would we need an individual email?

Fastjur:
Ah I see you found the issue for throwing the error too.

I have removed some unnecessary code too, we only need to add and remove by email and retrieve all subscriptions

Fastjur:
Woops. I fucked up the request param. But CH is closing now. Will fix when home :P

Fastjur:
So. To get back on the error messages. It appears that in the user controller we also simply send `ex.getMessage()`

How are these nice and readable errors, while they are not here?

src/main/java/ch/wisv/areafiftylan/extras/mailupdates/model/Subscription.java
```diff
@@ -0,0 +1,52 @@
...
+ * Used to gather emails from the frontend, to allow interested users
+ * to receive emails about the event.
+ *
+ * @author Jurriaan Den Toonder Created on 23-10-17
+ */
+@Entity
+@Data
+@RequiredArgsConstructor
+@NoArgsConstructor
+public class Subscription {
```
martijnjanssen: I just thought about this, it is very easy to allow users to unsubscribe, just add an active flag to the subscription.

Fastjur: Discussed, keeping it like this

Fastjur:
I have tested the migrations. They are working! *Ready to merge*

PR #446, opened 2017-10-27
Remove getUsername from User view

martijnjanssen:
Since we moved to the new login with the email field, the `getUsername` was still included in the `User` json view, this doesn't add anything but confusion, since the field is literally the same as the `email` field.


Fastjur:
Why aren't we just deprecating this completely? It is not used right?

skamoen:
It's part of Spring's `UserDetails` (see the `@Override`) so it has to be there.

Fastjur:
Ah check. I was not aware it is needed for the UserDetails.

In that case LGTM :+1:

PR #450, opened 2017-12-01
Scheduled banners

### LINKED ISSUE #268: Scheduled banner text
Banner texts should be able to be planned.

Examples:
- During week before tickets sale closes a deadline notification
- During event link to livestream
- After event a thank you message

elarb:
- `/web/banners` endpoint, to make it possible to schedule banners
- A `banner` has text and covers a date range in which the text should be displayed on the website
- Scheduling a new banner with a date range that is already (partially) covered by another banner is not possible (Note: this doesn't mean that conflicting date ranges cannot exists, since updates still can cause this)
- `/web/banners/current` returns the latest inserted scheduled banner where the current date (date when request was made) is within its date range

Closes #268

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
+++ b/src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
@@ -44,7 +44,7 @@ public class BannerController {
     @GetMapping("/current")
     public ResponseEntity<?> getCurrentBanner() {
         Banner currentBanner = bannerService.getCurrentbanner();
-        return new ResponseEntity<>(currentBanner, HttpStatus.OK);
+        return createResponseEntity(HttpStatus.OK, "Found a corresponding banner.", currentBanner);
     }
 
     /**
--- a/src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
+++ b/src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
@@ -1,0 +1,109 @@
+package ch.wisv.areafiftylan.web.banner.controller;
+
+import ch.wisv.areafiftylan.exception.BannerNotFoundException;
+import ch.wisv.areafiftylan.web.banner.model.Banner;
+import ch.wisv.areafiftylan.web.banner.service.BannerService;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.security.access.prepost.PreAuthorize;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.Collection;
+
+import static ch.wisv.areafiftylan.utils.ResponseEntityBuilder.createResponseEntity;
+
+@Controller
+@RequestMapping("/web/banners")
+public class BannerController {
+
+    private BannerService bannerService;
+
+    public BannerController(BannerService bannerService) {
+        this.bannerService = bannerService;
+    }
+
+    /**
+     * Get all banners in the database.
+     *
+     * @return all banners
+     */
+    @PreAuthorize("hasRole('ADMIN')")
+    @GetMapping
+    public Collection<Banner> getBanners() {
+        return bannerService.getAllBanners();
+    }
+
+    /**
+     * Returns a banner where the current date (date when the request was made)
+     * lies within the range of its start- and end-date.
+     *
+     * @return a banner
+     */
+    @GetMapping("/current")
+    public ResponseEntity<?> getCurrentBanner() {
+        Banner currentBanner = bannerService.getCurrentbanner();
+        return new ResponseEntity<>(currentBanner, HttpStatus.OK);
+    }
+
+    /**
+     * Adds a new banner to the database. The date-range should be valid (e.g. there shouldn't be
+     * another entity covering that range already).
+     *
+     * @param banner banner to be added
+     * @return a response-entity
+     */
+    @PreAuthorize("hasRole('ADMIN')")
+    @PostMapping
+    public ResponseEntity<?> addBanner(@RequestBody Banner banner) {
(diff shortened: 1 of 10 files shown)


src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Returns a banner where the current date (date when the request was made)
+     * lies within the range of its start- and end-date.
+     *
+     * @return a banner
+     */
+    @GetMapping("/current")
+    public ResponseEntity<?> getCurrentBanner() {
+        Banner currentBanner = bannerService.getCurrentbanner();
+        return new ResponseEntity<>(currentBanner, HttpStatus.OK);
```
MatthijsKok: We can use the createResponseEntity() helper method like you did below.

src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    @PreAuthorize("hasRole('ADMIN')")
+    @DeleteMapping
+    public ResponseEntity<?> deleteAllBanners() {
+        bannerService.deleteBanners();
+        return createResponseEntity(HttpStatus.OK,
+                "Successfully deleted all banners");
+    }
+
+    @ExceptionHandler(BannerNotFoundException.class)
```
MatthijsKok: What is the difference between handling here and in the centralized handling class?

src/test/java/ch/wisv/areafiftylan/integration/WebBannerIntegrationTest.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    @After
+    public void cleanUpBanners() {
+        bannerRepository.deleteAll();
+    }
+
+    //region GET all banners
+    @Test
+    public void testGetBannersWithoutPermissions() {
```
MatthijsKok: Could you please make sure that the naming of the tests is consistent with the rest of the repository? The current scheme is "...AsUser" or "...AsAdmin" instead of "...WithPermissions" etc...

src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
skamoen: The centralized class is for exceptions that occur in more than one controller. In this case, it's unlikely we'll be seeing a BannerNotFoundException outside of the BannerController so this is the correct placement.

skamoen:
I'm not 100% sure if `java.sql.Date` is the right type to use here. We've made some wrong assumptions about Hibernate compatibility before, but as this is a timing thing it's important timezones are included. Can you check if there's a `java.time` class available which works well with Hibernate and has a timezone?

src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
MatthijsKok: TIL, thanks 👍

elarb:
@skamoen Good point, I have looked into other variants before using `java.sql.Date`, but couldn't really find one that was appropriate enough. On the other hand, I don't think that timezones will change a lot in this case.

src/test/java/ch/wisv/areafiftylan/integration/WebBannerIntegrationTest.java
elarb: 👍

src/main/java/ch/wisv/areafiftylan/web/banner/controller/BannerController.java
elarb: Done

elarb:
@skamoen Still something you think should be different before merging?

src/main/java/ch/wisv/areafiftylan/web/banner/model/Banner.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Id
+    @GeneratedValue
+    private Long id;
+
+    private String text;
+
+    private Date startDate;
+
+    private Date endDate;
```
Fastjur: Wouldn't a LocalDateTime be much easier here?  
The `java.sql.date` class does not seem appropriate IMO.

src/main/java/ch/wisv/areafiftylan/web/banner/service/BannerServiceImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public Banner addBanner(Banner banner) {
+        return bannerRepository.save(banner);
+    }
+
+    @Override
+    public Banner update(Long bannerId, Banner banner) {
+        if (!bannerRepository.exists(bannerId))
```
skamoen: if without `{ }` 😱

Fastjur: It does work if it's a single line, you know that right :stuck_out_tongue:

skamoen: It works, but it's bad practice. There's no reason not to do it, and it prevents [nasty bugs](https://www.imperialviolet.org/2014/02/22/applebug.html)

Fastjur: Yeah it is a bit nasty yeah

elarb:
Since the frontend part on admin that depends on this PR already has been merged with the corresponding master branch, I think it would be convenient to merge this PR as well and eventually change stuff (like adding time next to date) later?

Fastjur:
I agree with that statement.

We can upgrade it later.

PR #455, opened 2017-12-17
Security package cleanup, fixes and tests

skamoen:
This PR removes some unused beans and lines from the security package. It fixes the logout endpoint (which never worked) and adds some tests to increase coverage to over 80%!


src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationController.java
```diff
@@ -82,14 +82,6 @@ public AuthenticationController(UserService userService, AuthenticationService a
         return createResponseEntity(HttpStatus.OK, "Token is valid!");
     }
 
-    @PreAuthorize("isAuthenticated()")
-    @PostMapping("/logout")
```
Fastjur: I'm not sure I understand anymore now.

We remove the logout postmapping. How does a user logout now?

skamoen: It's mapped to a native Spring Filter (as it's always been, this endpoint was never reachable) . See the security configuration

Fastjur: Ah I see. It makes sense to me now 👍

PR #457, opened 2017-12-23
Update sponsor controller

Fastjur:
This PR adds the following:

* Tests for deletion of a sponsor that is used under a tournament
* Tests for updating a sponsor
* Exception handler to return readable error message when trying to delete a sponsor that is under a tournament

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/web/sponsor/controller/SponsorController.java
+++ b/src/main/java/ch/wisv/areafiftylan/web/sponsor/controller/SponsorController.java
@@ -1,6 +1,6 @@
 package ch.wisv.areafiftylan.web.sponsor.controller;
 
-import ch.wisv.areafiftylan.exception.SponsorStillUsedInTournamentException;
+import ch.wisv.areafiftylan.exception.SponsorConstraintViolationException;
 import ch.wisv.areafiftylan.web.sponsor.model.Sponsor;
 import ch.wisv.areafiftylan.web.sponsor.model.SponsorType;
 import ch.wisv.areafiftylan.web.sponsor.service.SponsorService;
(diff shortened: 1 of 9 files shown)


src/main/java/ch/wisv/areafiftylan/web/sponsor/controller/SponsorController.java
```diff
@@ -78,4 +80,10 @@ public SponsorController(SponsorService sponsorService) {
         sponsorService.deleteAllSponsors();
         return createResponseEntity(HttpStatus.OK, "Successfully deleted all sponsors.");
     }
+
+    @ExceptionHandler(DataIntegrityViolationException.class)
+    public ResponseEntity<?> handleDataIntegrityViolationException(DataIntegrityViolationException ex) {
+        return createResponseEntity(HttpStatus.CONFLICT,
+                "Could not delete this sponsor because it is used by another entity!");
```
skamoen: This assumes only the DELETE call will trigger a `DataIntegrityViolation`. I get what you're trying to do, but this may lead to extemely confusing results in other cases. For example duplicate key entries will give the same exception but have nothing to do with deleting things. 

I recommend a response like "Could not complete request due to a constraint violation" or something. As the user calls this explicitly, it should make sense what the cause of the problem is.

Fastjur: Hmm that's a good point.

I might be able to walk through the root cause of this exception, check if it contains `could not delete` (or something similar), and then send a `HttpStatus.CONFLICT` with this message, and otherwise a `HttpStatus.BAD_REQUEST` such that the error message is generic in the frontend. Sounds good?

skamoen: That could work yes, try not to make it too hacky 😉

Fastjur:
I think the rename of the foreing key constraint requires a db migration right?

Fastjur:
image

This is how the error message now looks like when a sponsor still has tournaments linked to it.

src/main/java/ch/wisv/areafiftylan/web/sponsor/service/SponsorRepository.java
```diff
@@ -10,5 +10,7 @@
 @Repository
 public interface SponsorRepository extends JpaRepository<Sponsor, Long> {
 
+    Sponsor findSponsorById(Long id);
```
skamoen: This is already native to the JpaRepository as `findOne()`

PR #469, opened 2018-01-24
Use updated logo image url and properly format user

### LINKED ISSUE #467: Registration confirmation mail depending on non-existent user data
The confirmation mail received after registration starts with "Dear null null". We should replace it with something like "Dear user" instead.

### LINKED ISSUE #468: Image in mails doesn't exist
https://areafiftylan.nl/images-optimized/lancie/logo_without_date.png is the logo image used in mails, but its probably moved to another directory so it has to be updated

elarb:
- Update logo image URL in mail templates
- Properly format mail salutation

Closes #467 and closes #468

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/utils/mail/MailServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/utils/mail/MailServiceImpl.java
@@ -118,7 +118,7 @@ public class MailServiceImpl implements MailService {
         if (profile != null && profile.getFirstName() != null && profile.getLastName() != null) {
             return profile.getFirstName() + " " + profile.getLastName();
         }
-        return "User";
+        return "";
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


src/main/java/ch/wisv/areafiftylan/utils/mail/MailServiceImpl.java
```diff
@@ -113,11 +114,11 @@ private String prepareHtmlContent(String name, String message) {
...
     private String formatRecipient(User user) {
-        if (user.getProfile() != null) {
-            return user.getProfile().getFirstName() + " " + user.getProfile().getLastName();
-        } else {
-            return user.getEmail();
+        Profile profile = user.getProfile();
+        if (profile != null && profile.getFirstName() != null && profile.getLastName() != null) {
+            return profile.getFirstName() + " " + profile.getLastName();
         }
+        return "User";
```
skamoen: This sounds pretty dev-like. Let's either make this an email address, or turn it into something like

When defined:
>Hi First Last!

When undefined:
>Hi!

Should be easy to implement and avoids the awkward "Hi User"

elarb: Agreed

elarb: I prefer the latter (no mail). What would you say?

skamoen:
Could you also update https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/resources/templates/mailTemplate.html to "Hi" instead of "Dear"? I think "Dear" without a name is weird

PR #472, opened 2018-01-29
Possibly fix multiple authentication tokens bug

Fastjur:
This should prevent a user having multiple authentication tokens causing issues at login

Signed-off-by: Jurriaan Den Toonder <jur.den.toonder@gmail.com>


MatthijsKok:
Have you tested this? The default hibernate flush mode is [AUTO](https://docs.jboss.org/hibernate/orm/3.5/javadocs/org/hibernate/FlushMode.html), which ensures stale data is never returned. If I'm reading [link](http://docs.jboss.org/hibernate/core/4.3/manual/en-US/html/ch11.html#objectstate-flushing) correctly this change shouldn't matter. However since it is safer and more defensive I will merge it anyway.

PR #479, opened 2018-02-21
Add open and failed orderstatus to MolliePaymentService

skamoen:
There were some statuses missing from the PaymentService that were only now used (or noticed).
This PR adds two relevant missing statuses to the paymentservice and updates our status accordingly

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
@@ -55,7 +55,7 @@ public class MolliePaymentService implements PaymentService {
 
     @Override
     public String registerOrder(Order order) {
-        Client mollie =  new ClientBuilder().withApiKey(apiKey).build();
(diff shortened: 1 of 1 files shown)


src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java
```diff
@@ -113,10 +113,18 @@ public Order updateStatus(String orderReference) {
...
+                    case "open": {
+                        order.setStatus(OrderStatus.PENDING);
+                        break;
+                    }
                     case "cancelled": {
                         order.setStatus(OrderStatus.CANCELLED);
                         break;
                     }
+                    case "failed": {
+                        order.setStatus(OrderStatus.CANCELLED);
```
elarb: Why not create a new enum `FAILED`?

skamoen: Wouldn't add anything of value for us. The same logic applies when an order is cancelled.

PR #481, opened 2018-03-01
Export endpoint

### LINKED ISSUE #474: Login endpoint for external tournament planning system
Ask Paul for the details, I don't know the exact specifics, but with this, we don't have to make a database export before the event.

skamoen:
To make exporting and importing data easier for Paul, this adds an export endpoint containing the most essential fields about users and teams

Fixes #474

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
+++ b/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
@@ -68,6 +68,6 @@ public class Seat {
 
     @Override
     public String toString() {
-        return "" + seatGroup + seatNumber;
+        return seatGroup + seatNumber;
     }
 }
--- a/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
+++ b/src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
@@ -65,4 +65,9 @@ public class Seat {
     public boolean isTaken() {
         return ticket != null;
     }
+
+    @Override
+    public String toString() {
+        return "" + seatGroup + seatNumber;
+    }
 }
--- a/src/main/java/ch/wisv/areafiftylan/users/model/UserExportDTO.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/model/UserExportDTO.java
@@ -11,7 +11,8 @@ import java.util.List;
 public class UserExportDTO {
 
     private String email;
-    private String password_hash;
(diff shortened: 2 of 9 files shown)


src/main/java/ch/wisv/areafiftylan/seats/model/Seat.java
```diff
@@ -65,4 +65,9 @@ public User getUser() {
     public boolean isTaken() {
         return ticket != null;
     }
+
+    @Override
+    public String toString() {
+        return "" + seatGroup + seatNumber;
```
martijnjanssen: This looks hideous, please change this.

src/main/java/ch/wisv/areafiftylan/users/model/UserExportDTO.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+
+@Getter
+@AllArgsConstructor
+public class UserExportDTO {
+
+    private String email;
+    private String password_hash;
+    private Long userId;
+    private List<Seat> seat;
```
martijnjanssen: Why not remove this here and add it as a separate export? An export to a `SeatExportDTO` with `Long userId` and `String seatString`?

src/main/java/ch/wisv/areafiftylan/utils/ExportController.java
```diff
@@ -0,0 +1,49 @@
...
+    private final TicketService ticketService;
+
+    public ExportController(SeatService seatService, TeamService teamService, TicketService ticketService) {
+        this.seatService = seatService;
+        this.teamService = teamService;
+        this.ticketService = ticketService;
+    }
+
+    @GetMapping
+    public Map<String, ?> exportUsers() {
```
martijnjanssen: This function looks horribly formatted here, that needs to improve.

src/main/java/ch/wisv/areafiftylan/users/model/UserExportDTO.java
skamoen: A User has a seat, why would you export that separately. It's all tied to the userId, which is the whole point of this endpoint right?

martijnjanssen: True, but if we want to export more, it's more expandable. It looks a little sloppy and more thrown together now. We could use it this way, but it looks a bit low quality :stuck_out_tongue:

PR #483, opened 2018-03-02
Fix Export endpoint nullpointer

skamoen:
The testdata didn't cover the possible scenario's we have in the live database, so the export endpoint resulted in a nullpointer.

This PR expands the export method for better null checking


martijnjanssen:
Okay this looks better, perfect!

PR #493, opened 2018-09-29
Token persistence issue fixes

### LINKED ISSUE #473: Startup task to clear all login tokens

### LINKED ISSUE #484: Unable to delete team in some cases
We had an issue with a user not being able to delete his team, while he was the captain. Apparently, the invites that were already `used`, stay in the database, causing a constraint violation when trying to delete the team.

- We should delete an invite token from the database when it is used, there is no reason to keep it there.
- Or, we should delete all invite tokens, `used(TRUE/FALSE)` when deleting a team.

skamoen:
This PR fixes two minor token-related issues.

First issue, #473, extremely hard to reproduce but to make sure no conflicts occur between restarts, we delete all auth tokens on startup. This means all users will need to login after a new deploy.

Second issue #484 was caused by a cascading operation, as we didn't delete invites on revoke. When we wanted to delete a team, all previous invites were still persisted, which would result in orphaned invites. This PR removes all invites on deletion (and adds a Transactional annotation in case either operation fails)

Fixes #484 
Fixes #473

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
@@ -24,10 +24,12 @@ import ch.wisv.areafiftylan.security.token.repository.AuthenticationTokenReposit
 import ch.wisv.areafiftylan.users.model.User;
 import ch.wisv.areafiftylan.users.service.UserService;
 import com.google.common.base.Strings;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service
+@Slf4j
 public class AuthenticationServiceImpl implements AuthenticationService {
 
     private final AuthenticationTokenRepository authenticationTokenRepository;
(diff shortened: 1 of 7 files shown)


src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -68,4 +70,10 @@ public void removeAuthToken(String xAuth) {
         token.revoke();
         authenticationTokenRepository.saveAndFlush(token);
     }
+
+    @Override
+    public void removeAllAuthTokens() {
+        authenticationTokenRepository.deleteAll();
```
martijnjanssen: Shouldn't we `revoke()` all non-revoked tokens instead of dropping database contents? Seems like this is clearing log-in history, which we may not want.

skamoen: I don't think we should rely on persisted authenticationTokens for login history. If we want to keep tabs on that, this would be the place to log it: https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/security/JsonLoginAuthenticationAttemptHandler.java#L49

PR #497, opened 2018-10-06
Add missing import of no longer bundled jaxb

martijnjanssen:


build.gradle
```diff
@@ -18,9 +18,7 @@ apply plugin: 'io.spring.dependency-management'
 apply plugin: 'jacoco'
 
 jar {
-    baseName = 'lancie-api'
-    version = '2017.1.6'
```
TimvdLippe: Should we really remove this?

martijnjanssen: The jar file is only used in the container, and the container originates from a specific build with a specific tag. I think we no longer need it. This was needed when there were multiple jar files in one archive directory on the deployment server.

martijnjanssen: We could change to this format? But I'm not quite sure when it's needed.

https://github.com/WISVCH/events/blob/develop/build.gradle#L24-L26

skamoen:
For reference: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-with-Java-9-and-above#jaxb

PR #500, opened 2018-10-08
Gradle improvements

MatthijsKok:


build.gradle
```diff
@@ -69,19 +64,10 @@ eclipse {
     }
 }
 
-task wrapper(type: Wrapper) {
-    gradleVersion = '4.4'
+wrapper {
+    gradleVersion = '4.10'
 }
 
-//task bootRunDev(type: BootRunTask) {
```
TimvdLippe: This is useful for frontend developers. Why was it removed? :(

martijnjanssen: It no longer worked, but maybe @MatthijsKok can find a way to get it to work again? In the meantime, we have `export SPRING_PROFILES_ACTIVE=dev; ./gradlew bootRun`.

PR #512, opened 2020-01-07
Order Expiration logging and tests

skamoen:
This PR adds some logging to the Expiration process to get some insight into possible errors. It also adds a test for the scheduled task responsible for the proces.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
@@ -74,10 +74,6 @@ public class AuthenticationServiceImpl implements AuthenticationService {
         AuthenticationToken token =
                 authenticationTokenRepository.findByToken(xAuth).orElseThrow(XAuthTokenNotFoundException::new);
 
-        if (!token.isValid()) {
-            throw new InvalidTokenException();
-        }
-
         token.revoke();
         authenticationTokenRepository.saveAndFlush(token);
     }
--- a/src/main/java/ch/wisv/areafiftylan/utils/TaskScheduler.java
+++ b/src/main/java/ch/wisv/areafiftylan/utils/TaskScheduler.java
@@ -25,6 +25,7 @@ import ch.wisv.areafiftylan.security.authentication.AuthenticationService;
 import ch.wisv.areafiftylan.security.token.VerificationToken;
 import ch.wisv.areafiftylan.security.token.repository.VerificationTokenRepository;
 import ch.wisv.areafiftylan.users.service.UserRepository;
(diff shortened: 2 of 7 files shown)


src/main/java/ch/wisv/areafiftylan/utils/TaskScheduler.java
```diff
@@ -75,8 +77,10 @@ public void ExpireOrders() {
         Collection<Order> allOrdersBeforeDate = orderRepository.findAllByCreationDateTimeBefore(expireBeforeDate);
 
         List<Order> expiredOrders = allOrdersBeforeDate.stream().filter(isExpired()).collect(Collectors.toList());
-
-        expiredOrders.forEach(orderService::expireOrder);
+        if (expiredOrders.size() > 0) {
```
martijnjanssen: Is there no empty check on a list?

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -74,10 +74,6 @@ public void removeAuthToken(String xAuth) {
         AuthenticationToken token =
                 authenticationTokenRepository.findByToken(xAuth).orElseThrow(XAuthTokenNotFoundException::new);
 
-        if (!token.isValid()) {
```
martijnjanssen: Why is this removed?

skamoen:
It looks like the TaskScheduler isn't initialized on startup... Let's not merge this yet.

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
skamoen: This just sets a boolean. The thrown exception here causes a bunch of pointless stacktraces in the log. Worst case, you set the boolean to false twice. Best case, no more stacktraces.

martijnjanssen: If the token is already not valid, why not just `return` here? Saves a write to the database and maybe sideeffects from calling `revoke()` again.

skamoen:
Apparently there was a naming conflict with TaskScheduler in a newer Spring version. It worked in the test because of an explicit Autowired field. After the rename, it's initialized as a proper component

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
skamoen: There could be different reasons for an invalid token. 

`isUsed() || isExpired() || isRevoked()`

An explicit revoke seems like a good idea here, even if we don't have a direct use now.

PR #518, opened 2020-02-25
Fix hardcoded pickup

bramvankooten:


skamoen:
Let's fix this properly in #519

PR #525, opened 2020-03-03
Bug add birthday change constraint

### LINKED ISSUE #520: Add birthday change constraint for checked in users to edit profile
The frontend only uses the `users/current/profile` call, but the api exposes another endpoint which does not have the check for limiting birthday changes when a user is checked in.

https://github.com/AreaFiftyLAN/lancie-api/commit/378258b3c4ada8d2c776ea50e05afc48641de68a#diff-69bbd75df2595c3791120aea66696788R107

julian9499:
fixes #520  
The issue was that the function compared date objects instead of the dates as string. Which resulted in the tests working since we used the same object, but the real life requests not. This has been solved in this pull request. The function has also been put in a separate method which allows the protection of all profile edit paths.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
@@ -55,7 +55,7 @@ public class UserProfileRestController {
      * @param input The changes
      * @return boolean whether they are allowed to edit their profile
      */
-    private boolean checkAllowedToEditProfile(User user, ProfileDTO input) {
+    private boolean allowedToEditProfile(User user, ProfileDTO input) {
         boolean isUserCheckedIn = rfidService.isOwnerLinked(user.getEmail());
         LocalDate currentBirthday = user.getProfile().getBirthday();
         boolean isDateChanged = !currentBirthday.equals(input.getBirthday());
@@ -78,11 +78,11 @@ public class UserProfileRestController {
     @PreAuthorize("isAuthenticated()")
     @PostMapping("/current/profile")
     public ResponseEntity<?> addProfile(@AuthenticationPrincipal User user, @Validated @RequestBody ProfileDTO input) {
-        if (checkAllowedToEditProfile(user, input)) {
+        if (allowedToEditProfile(user, input)) {
             User changedUser = userService.addProfile(user.getId(), input);
             return createResponseEntity(HttpStatus.OK, "Profile successfully changed", changedUser.getProfile());
         } else {
-            return createResponseEntity(HttpStatus.BAD_REQUEST, "Unable to change date during event", user.getProfile());
+            return createResponseEntity(HttpStatus.BAD_REQUEST, "Not permitted to change date during event", user.getProfile());
         }
     }
 
--- a/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
@@ -85,7 +85,7 @@ public class UserProfileRestController {
         }
 
         LocalDate currentBirthday = user.getProfile().getBirthday();
-        boolean isDateChanged = currentBirthday != input.getBirthday();
+        boolean isDateChanged = !currentBirthday.toString().equals(input.getBirthday().toString());
 
         // If rfidLinks are present and the date is changed then return an error
         if (isDateChanged) {
--- a/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
@@ -64,6 +64,22 @@ public class UserProfileRestController {
     }
 
     /**
+     * An addProfile function that returns a custom message used during the event.
+     *
+     * Add a profile to a user. An empty profile is created when a user is created, so this method
+     * fills the existing fields
+     *
+     * @param userId The userId of the user to which the profile needs to be added
+     * @param input  A representation of the profile
+     * @return The user with the new profile
+     */
+    public ResponseEntity<?> addProfileDuringEvent(Long userId, ProfileDTO input) {
+        User user = userService.addProfile(userId, input);
+
+        return createResponseEntity(HttpStatus.OK, "Unable to change date during event. The rest of profile successfully set", user.getProfile());
+    }
+
+    /**
      * Add a profile to the current user. An empty profile is created when a user is created, so
      * this method fills the existing fields.
      * <p>
--- a/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
+++ b/src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
@@ -46,11 +46,20 @@ public class UserProfileRestController {
         this.userService = userService;
         this.rfidService = rfidService;
     }
-    private ResponseEntity<?> checkBeforeAddingProfile(User user, ProfileDTO input){
+
+    /**
+     * This method returns a boolean indicating whether an user is allowed to make the changes to
+     * their profile or not.
+     *
+     * @param user The concerning user
+     * @param input The changes
+     * @return boolean whether they are allowed to edit their profile
+     */
+    private boolean checkAllowedToEditProfile(User user, ProfileDTO input) {
         boolean isUserCheckedIn = rfidService.isOwnerLinked(user.getEmail());
 
         if (!isUserCheckedIn) {
-            return this.addProfile(user.getId(), input);
+            return true;
         }
 
         LocalDate currentBirthday = user.getProfile().getBirthday();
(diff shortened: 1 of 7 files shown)


src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
```diff
@@ -48,19 +48,29 @@
...
+            return true;
+        }
+
+        LocalDate currentBirthday = user.getProfile().getBirthday();
+        boolean isDateChanged = !currentBirthday.toString().equals(input.getBirthday().toString());
+
+        // If rfidLinks are present and the date is changed then return an error
+        if (isDateChanged) {
+            return false;
+        }
```
skamoen: You can inline all these variables

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
```diff
@@ -48,19 +48,29 @@
...
-        User user = userService.addProfile(userId, input);
+    private boolean checkAllowedToEditProfile(User user, ProfileDTO input) {
+        boolean isUserCheckedIn = rfidService.isOwnerLinked(user.getEmail());
+
+        if (!isUserCheckedIn) {
+            return true;
+        }
+
+        LocalDate currentBirthday = user.getProfile().getBirthday();
+        boolean isDateChanged = !currentBirthday.toString().equals(input.getBirthday().toString());
```
skamoen: Why would you do an equals on the `toString()` methods and not just the objects itself?

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
```diff
@@ -48,19 +48,29 @@
...
+     * @return boolean whether they are allowed to edit their profile
      */
-    @PreAuthorize("@currentUserServiceImpl.canAccessUser(principal, #userId)")
-    @PostMapping("/{userId}/profile")
-    public ResponseEntity<?> addProfile(@PathVariable Long userId, @Validated @RequestBody ProfileDTO input) {
-        User user = userService.addProfile(userId, input);
+    private boolean checkAllowedToEditProfile(User user, ProfileDTO input) {
+        boolean isUserCheckedIn = rfidService.isOwnerLinked(user.getEmail());
+
+        if (!isUserCheckedIn) {
```
skamoen: Just inline this variable

src/test/java/ch/wisv/areafiftylan/integration/UserRestIntegrationTest.java
```diff
@@ -25,9 +25,11 @@
 import io.restassured.http.ContentType;
 import io.restassured.http.Header;
 import org.apache.http.HttpStatus;
+import org.apache.tomcat.jni.Local;
```
skamoen: What's this?

julian9499: I don't know. Probably a wrong import it was unused. Removed it!

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
julian9499: Done!

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
julian9499: Removed it and changed it to use the equals method

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
julian9499: Done!

src/main/java/ch/wisv/areafiftylan/users/model/Profile.java
```diff
@@ -82,4 +82,8 @@ public void setAllFields(String firstName, String lastName, String displayName,
         this.phoneNumber = phoneNumber;
         this.notes = notes;
     }
+
+    public void setBirthday(LocalDate birthday){
```
Fastjur: Can use lombok for getters and setters

src/main/java/ch/wisv/areafiftylan/users/model/User.java
```diff
@@ -100,6 +101,7 @@ public String getPassword() {
 
     public void resetProfile() {
         this.profile = new Profile();
+        this.profile.setBirthday(LocalDate.now());
```
Fastjur: If we reset a profile, should we set the birthdate to now, or to undefined/null ?

It would make more sense to clear it instead of having an arbitrary date filled in.
If that raises errors in the alcohol check, just check for null first.

julian9499: The profile has @nonNull tho

src/main/java/ch/wisv/areafiftylan/users/model/Profile.java
julian9499: done!

src/main/java/ch/wisv/areafiftylan/users/model/User.java
julian9499: I will fill in a arbritrary date

Fastjur: Is it not possible for someone to not have a birthdate then? What about the case where ticket transfers occur?

julian9499: I really don't know enough about the ticket transfers to answer this

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
```diff
@@ -77,22 +78,28 @@
...
-        LocalDate currentBirthday = user.getProfile().getBirthday();
-        boolean isDateChanged = currentBirthday != input.getBirthday();
-
-        // If rfidLinks are present and the date is changed then return an error
-        if (isDateChanged) {
+        if (checkAllowedToEditProfile(user, input)) {
+            User changedUser = userService.addProfile(user.getId(), input);
+            return createResponseEntity(HttpStatus.OK, "Profile successfully changed", changedUser.getProfile());
+        } else {
             return createResponseEntity(HttpStatus.BAD_REQUEST, "Unable to change date during event", user.getProfile());
```
martijnjanssen: Not permitted maybe? The request itself is valid.

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
```diff
@@ -77,22 +78,28 @@
...
-        if (!isUserCheckedIn) {
-            return this.addProfile(user.getId(), input);
-        }
-
-        LocalDate currentBirthday = user.getProfile().getBirthday();
-        boolean isDateChanged = currentBirthday != input.getBirthday();
-
-        // If rfidLinks are present and the date is changed then return an error
-        if (isDateChanged) {
+        if (checkAllowedToEditProfile(user, input)) {
```
martijnjanssen: Just allowedToEditProfile?

src/main/java/ch/wisv/areafiftylan/users/model/User.java
```diff
@@ -100,6 +101,7 @@ public String getPassword() {
 
     public void resetProfile() {
         this.profile = new Profile();
+        this.profile.setBirthday(LocalDate.parse("2000-01-01"));
```
martijnjanssen: This way you're automatically older than 18 after resetting your profile.

Fastjur: Yes this was my thought. I would not set this to any date tbh. Maybe the birthday should be nullable then?

Fastjur: Maybe localdate.now() makes more sense after all? As @julian9499 implemented initially?

src/main/java/ch/wisv/areafiftylan/users/model/Profile.java
```diff
@@ -31,7 +31,7 @@
 
 @Entity
 @Data
-@NoArgsConstructor
+    @NoArgsConstructor
```
Fastjur: Whitespace

julian9499: fixed!

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
julian9499: Done!

src/main/java/ch/wisv/areafiftylan/users/controller/UserProfileRestController.java
julian9499: Done!

src/main/java/ch/wisv/areafiftylan/users/model/User.java
julian9499: Done that @martijnjanssen is this oke?

PR #532, opened 2020-12-17
JUnit5 test migration

skamoen:


martijnjanssen (approved):
All except maybe 1-2 seem to be fixed with a replace, looks good!

PR #537, opened 2020-12-23
Allow up to 5 sessions for users

DevDonut:
Allow up to 5 sessions for users, so you won't be logged out of the admin site when logging in to the regular site and vice versa.

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
@@ -64,7 +64,7 @@ public class AuthenticationServiceImpl implements AuthenticationService {
         // Delete the old Tokens if there are more than 4.
         List<AuthenticationToken> existingTokens = authenticationTokenRepository.findByUserEmail(email);
         if (existingTokens.size() >= 4) {
-            authenticationTokenRepository.deleteAll(existingTokens);
+            authenticationTokenRepository.delete(existingTokens.get(0));
         }
 
         return authenticationTokenRepository.saveAndFlush(new AuthenticationToken(user)).getToken();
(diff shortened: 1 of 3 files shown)


skamoen:
An Optional of a List is is bit much. If you omit the Optional, a query with no results will just give you an empty list, making the whole thing a lot easier.

dsluijk:
This feels like an hack, it might be better to replace it with something like JWT tokens? Should be lighter on the database too.

DevDonut:
> This feels like an hack, it might be better to replace it with something like JWT tokens? Should be lighter on the database too.

@dsluijk i feel like this would require a complete overhaul of the Authentication service as currently implemented. I myself have no experience with JWT tokens, so I'm not sure how much work that would be.

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
```diff
@@ -58,8 +61,11 @@ public AuthenticationServiceImpl(AuthenticationTokenRepository authenticationTok
     public String createNewAuthToken(String email) {
         User user = userService.getUserByEmail(email);
 
-        // Delete the old Token
-        authenticationTokenRepository.findByUserEmail(email).ifPresent(authenticationTokenRepository::delete);
+        // Delete the old Tokens if there are more than 4.
+        List<AuthenticationToken> existingTokens = authenticationTokenRepository.findByUserEmail(email);
+        if (existingTokens.size() >= 4) {
+            authenticationTokenRepository.deleteAll(existingTokens);
```
martijnjanssen, line 67: So if there are more than 4 tokens, we're deleting them all? Makes more sense to only delete the oldest right? Deleting them all introduces a sort of random factor for the users. One time they are still logged in, and the other time all devices are signed out.

martijnjanssen (commented):
Cool feature! This is quite the improvement

martijnjanssen:
> This feels like an hack, it might be better to replace it with something like JWT tokens? Should be lighter on the database too.

This is definitely not a hack, the solution proposed is reasonable. Switching to JWT also has many downsides (e.g. elaborate checking logic for users that still might have valid token while the user is invalid/banned), apart from the fact that re-implementing the authentication makes little sense.

src/main/java/ch/wisv/areafiftylan/security/authentication/AuthenticationServiceImpl.java
DevDonut: That's a solid point! I will change it.

skamoen (approved):

PR #540, opened 2021-01-02
Return correct status codes and messages when a user changes their password

DevDonut:


dsluijk (approved):

src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
```diff
@@ -89,8 +90,17 @@
     @PostMapping("/password")
     public ResponseEntity<?> changeCurrentUserPassword(@AuthenticationPrincipal User user,
                                                        @RequestBody @Validated PasswordChangeDTO passwordChangeDTO) {
-        userService.changePassword(user.getId(),
-                passwordChangeDTO.getOldPassword(), passwordChangeDTO.getNewPassword());
+        try {
+            userService.changePassword(user.getId(),
+                    passwordChangeDTO.getOldPassword(), passwordChangeDTO.getNewPassword());
+        } catch (Exception e) {
```
dsluijk, line 96: You don't catch any other errors than the two listed, which would then be silently ignored. Maybe better to add a simple case with returns a 500 error?

dsluijk (changes requested):

src/main/java/ch/wisv/areafiftylan/users/controller/CurrentUserRestController.java
DevDonut: I looked at the errors it could throw, in principle it only throws the two listed, but I agree with adding the last case

dsluijk (approved):

ArdyZ:
@dsluijk and @DevDonut what should be done to fix this pull request and merge it?

Issue #542, opened 2021-01-12
CORS errors on some endpoints


dsluijk:
We are getting CORS errors on some endpoints. This might be on all PUT requests, but this has to be verified. As the impact is unknown we should look into it soon.

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://api.areafiftylan.nl/api/v1/tickets/tickettype/5186. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://api.areafiftylan.nl/api/v1/tickets/tickettype/5186. (Reason: CORS request did not succeed).
```

skamoen:
You need to add allowed methods here: https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java#L126

It seems by default only GET, POST and OPTIONS are allowed

dsluijk:
Solved

PR #543, opened 2021-01-13
Allow Cross-Origin PUT Requests

DevDonut:


skamoen (changes requested):
You're missing the DELETE method. Alternatively, you can set it to `"*"` to allow all methods

dsluijk:
Closes #542.

I think changing it to `*` is the best way, and we don't have restricted API calls reliant on CORS.

src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
```diff
@@ -123,7 +123,7 @@ public WebMvcConfigurer corsConfigurer() {
         return new WebMvcConfigurer() {
             @Override
             public void addCorsMappings(@Nonnull CorsRegistry registry) {
-                registry.addMapping("/**").exposedHeaders("X-Auth-Token");
+                registry.addMapping("/**").exposedHeaders("X-Auth-Token").allowedMethods("GET", "POST", "PUT", "OPTIONS", "HEAD");
```
dsluijk, line 126:
```suggestion
                registry.addMapping("/**").exposedHeaders("X-Auth-Token").allowedMethods("*");
```

dsluijk (changes requested):

src/main/java/ch/wisv/areafiftylan/security/SecurityConfiguration.java
DevDonut: I wasn't sure whether this would cause any security concerns, but if you don't think so, I'll change it!

dsluijk: No, this does not change any actual security policies of the server, just requests from other domains (cross origin). As main site (https://areafiftylan.nl) is on a different origin than the api (https://api.areafiftylan.nl) we are required to add this for the site to work properly.

dsluijk (approved):

PR #551, opened 2021-12-24
Making a function to allow free tickets

### LINKED ISSUE #550: Make it possible to buy tickets with a total €0.00
At the moment we cannot have free tickets (e.g. PreLAN is free). Therefore the following should happen:
- In [OrderServiceImpl.java](https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java#L207)  the `requestPayment` should have another case.
- This case should set the order to "PAID"
- This case should return the following url: https://areafiftylan.nl/order-check and maybe add the ?order=<orderId> behind it.

This should be done after checking if the order is empty, to prevent unwanted problems

ArdyZ:
This function allows having a total ticket price of 0.00 euro. It will directly set the status to PAID and move to the "https://areafiftylan.nl/order-check?order=" + orderId url.

closes #550

### DIFF
--- a/src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
@@ -93,6 +93,11 @@ public class Order {
         return price;
     }
 
+    @JsonView(View.OrderOverview.class)
+    public float getAmountTickets() {
+        return tickets.size();
+    }
+
     public void setUser(User user) {
         this.user = user;
         if (user != null) {
--- a/src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
@@ -94,7 +94,7 @@ public class Order {
     }
 
     @JsonView(View.OrderOverview.class)
-    public float getAmountTickets() {
+    public int getAmountTickets() {
         return tickets.size();
     }
 
--- a/src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
+++ b/src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
@@ -52,6 +52,9 @@ public class OrderServiceImpl implements OrderService {
     private final PaymentService paymentService;
     private final MailService mailService;
 
+    @Value("${a5l.paymentReturnUrl}")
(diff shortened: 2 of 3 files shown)


src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
```diff
@@ -93,6 +93,11 @@ public float getAmount() {
         return price;
     }
 
+    @JsonView(View.OrderOverview.class)
+    public float getAmountTickets() {
```
skamoen: Why is this a `float`? `Integer` or `int` seems more suitable

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
```diff
@@ -206,12 +206,16 @@ public Order removeTicketFromOrder(Long orderId, Long ticketId) {
...
-        if (order.getAmount() == 0) {
+        if (order.getAmountTickets() == 0) {
             throw new IllegalStateException("Order can not be empty");
         }
         if (order.getStatus() != OrderStatus.ASSIGNED) {
             throw new UnassignedOrderException(order.getId());
         }
+        if (order.getAmount() == 0) {
+            order.setStatus(OrderStatus.PAID);
+            return "https://areafiftylan.nl/order-check?order=" + orderId;
```
skamoen, line 217: Please build this url with a String from a property. See https://github.com/AreaFiftyLAN/lancie-api/blob/master/src/main/java/ch/wisv/areafiftylan/products/service/MolliePaymentService.java#L45-L49 for an example. Maybe you can reuse a property, or else add a new one

skamoen (changes requested):

src/main/java/ch/wisv/areafiftylan/products/model/order/Order.java
ArdyZ: Fixed

src/main/java/ch/wisv/areafiftylan/products/service/OrderServiceImpl.java
ArdyZ: Fixed

skamoen (approved):

PR #557, opened 2022-02-24
Use ghcr and updated github actions instead of quay

julian9499:


Fastjur:
I suggest not merging this until after the pre-lan event tomorrow.

julian9499:
If this goes wrong it's on my head :shipit:

Fastjur:
Alright, we can always revert pretty easily ;)

Fastjur (approved):

PR #562, opened 2022-03-01
Ticket transfer endpoint

ArdyZ:


ArdyZ:
@skamoen could you check out if this is the correct way of doing it?

